PR #5, opened 2013-03-27
Add a 'masked' extension to handle masked input

wking:
This doesn't actually seem to be substituting MaskedCovarianceMatrix
instances like it's supposed to, but I thought I'd post my work in
progress so others could chime in.

See the associated mailing list discussion for more details:

http://sourceforge.net/mailarchive/forum.php?thread_name=20130325121945.GI18933%40bio230.biologie.hu-berlin.de&forum_name=mdp-toolkit-users

### DIFF
--- a/mdp/__init__.py
+++ b/mdp/__init__.py
@@ -164,6 +164,7 @@ from classifier_node import (ClassifierNode, ClassifierCumulator)
 # import our modules
 import nodes
 import hinet
+import masked
 import parallel
 from test import test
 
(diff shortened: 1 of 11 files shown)


otizonaizit:
I assume nothing is going to happen with this PR, right? I would close it in a couple of days, unless the original author complains :)

wking:
Yeah, close away.  Sorry I haven't had time to finish it up :/.

PR #15, opened 2015-10-20
NEW: Add function to generate a random correlation matrix

philipdeboer:
Let me know whether the function signatures are intuitive and if there are any obvious performance improvements that we should make.


otizonaizit:
This has been moved to scipy, right? Could you link here the PR to scipy (or he commit where this has been merged)? After that I would close the PR. Thanks!

philipdeboer:
Yes, see https://github.com/scipy/scipy/pull/5859.

PR #16, opened 2016-03-07
BF: workaround for exec not being parsed (?) correctly on python 2.7.3

yarikoptic:

PR #17, opened 2016-03-07
Add wrapper function for exec

keszybz:
Old Python versions don't allow exec statement in nested functions:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "mdp/__init__.py", line 131, in <module>
    from . import utils
  File "mdp/utils/__init__.py", line 29, in <module>
    from .slideshow import (basic_css, slideshow_css, HTMLSlideShow,
  File "mdp/utils/slideshow.py", line 23, in <module>
    from . import templet
  File "mdp/utils/templet.py", line 196
    exec(code, globals, kw)
SyntaxError: unqualified exec is not allowed in function 'expand' it is a nested function
```

Different solution for #16.

### DIFF
--- a/mdp/utils/templet.py
+++ b/mdp/utils/templet.py
@@ -179,6 +179,8 @@ class _TemplateBuilder(object):
                     code.append(s + self.emitpat % part)
         return '\n'.join(code)
 
+def _exec(code, globals, locals):
+    exec(code, globals, locals)
 
 class _TemplateMetaClass(type):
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
that is just .... nice ;)  seems to work!  I again wish there was at least some rudimentary .travis.yml for mdp... then we could know "for sure" without trying ;)

keszybz:
Yeah, a .travis.yml file would be useful :) My travis-foo is very low though.

PR #18, opened 2016-03-08
some rudimentary travis setup

yarikoptic:
Let's see if it would work

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+# vim ft=yaml
+# travis-ci.org definition for MDP build (based on PyMVPA configuration
+# which in turn was based on nipype and nipy)
+#
+# We pretend to be erlang because we need can't use the python support in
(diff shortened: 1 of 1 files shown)


yarikoptic:
Please on your end go to travis-ci.org , login through github.com and enable testing on travis for your repo.

Do the same on coveralls.io and codecov.io -- both are used for coverage reporting ;)

yarikoptic:
and it works until getting stuck, see https://travis-ci.org/yarikoptic/mdp-toolkit/builds/114428551

pberkes:
Why was it merged? the run did not succeed

otizonaizit:
to have something to work on. I fixed it already ;)

Debilski:
Piping in some http request does not encourage local testing of the travis file, though.

keszybz:
Wow, not only native Python 3 support but now even CI. Thanks.

PR #20, opened 2017-01-18
Code review

varunrajk:
Hi MDP Developers, 

Here is my first pull request. 

It includes only an OnlineNode, OnlineFlow and CircularFlow. I added corresponding test functions for each and also two online nodes (Moving average and Moving time diff) as basic online nodes. 

Looking forward to your edits and comments!

-Varun

### DIFF
--- a/mdp/linear_flows_online.py
+++ b/mdp/linear_flows_online.py
@@ -11,22 +11,50 @@ class OnlineFlowException(mdp.MDPException):
 
 
 class OnlineFlow(mdp.Flow):
-    """An 'OnlineFlow' is a sequence of online/executable nodes that are trained and executed
+    """An 'OnlineFlow' is a sequence of nodes that are trained online and executed
     together to form a more complex algorithm.  Input data is sent to the
     first node and is successively processed by the subsequent nodes along
     the sequence.
 
-    Using an online flow as opposed to handling manually a set of nodes has a
-    clear advantage: The general online flow implementation automatizes the
+    Using an online flow as opposed to manually handling a set of nodes has a
+    clear advantage: The general online flow implementation automatates the
     training (including supervised training and multiple training phases),
     execution, and inverse execution (if defined) of the whole sequence.
 
-    OnlineFlow sequence can contain trained or non-trainable Nodes and optionally
-    a terminal trainable Node.
+    OnlineFlow sequence can contain
+    (a) only OnlineNodes, or
+    (b) a mix of OnlineNodes and trained or non-trainable Nodes (classic), or
+    (c) a mix of OnlineNodes/trained or non-trainable Nodes and a terminal trainable Node whose
+    training hasn't finished.
+
+    Here is a brief summary of the terminology used:
+       A "trainable" node: node.is_trainable() returns True, node.is_training() returns True.
+       A "trained" node: node.is_trainable() returns True, node.is_training() returns False.
+       A "non-trainalbe" node: node.is_trainable() returns False, node.is_training() returns False.
 
     Differences between a Flow and an OnlineFlow:
-    a) In Flow, data is processed sequentially, training one node at a time. Whereas, in an
-       OnlineFlow data is processed simultaneously training all the nodes at the same time.
+    a) In Flow, data is processed sequentially, training one node at a time. That is, the second
+       node's training starts only after the first node is "trained". Whereas, in an OnlineFlow data is
+       processed simultaneously training all the nodes at the same time.
+       Eg:
+
+       flow = Flow([node1, node2]), onlineflow = OnlineFlow([node1, node2])
+
+       Let input x = [x_0, x_1, ...., x_n], where x_t a sample or a mini batch of samples.
+
+       Flow training:
+            node1 trains on the entire x. While node1 is training, node2 is inactive.
+            node1 training completes. node2 training begins on the node1(x).
+
+            Therefore, Flow goes through all the data twice. Once for each node.
+
+       OnlineFlow training:
+            node1 trains on x_0. node2 trains on the output of node1 (node1(x_0))
+            node1 trains on x_1. node2 trains on the output of node1 (node1(x_1))
+            ....
+            node1 trains on x_n. node2 trains on the output of node1 (node1(x_n))
+
+            OnlineFlow goes through all the data only once.
 
     b) Flow requires a list of dataiterables with a length equal to the
        number of nodes or a single numpy array. OnlineFlow requires only one
@@ -37,7 +65,7 @@ class OnlineFlow(mdp.Flow):
        that contains tuples of args for each node: [x, (node0 args), (node1 args), ...]. See
        train docstring.
 
-    d) OnlineFlow also has the cache attribute that retrieves any caches variables
+    d) OnlineFlow also has the cache attribute that retrieves any temporary caches variables
        stored in the OnlineNodes.
 
     Crash recovery is optionally available: in case of failure the current
@@ -177,7 +205,7 @@ class OnlineFlow(mdp.Flow):
         in batches (=shape[1]).
 
         'data_iterables' can also return a list or a tuple, where the first entry is
-        'x' and the following are the required args for training all the nodes in
+        'x' and the rest are the required args for training all the nodes in
         the flow (e.g. for supervised training).
 
         (x, (node-0 args), (node-1 args), ..., (node-n args)) - args for n nodes
@@ -233,7 +261,6 @@ class OnlineFlow(mdp.Flow):
         else:
             self._check_value_type_is_compatible(value)
 
-        # make a copy of list
         flow_copy = list(self.flow)
         flow_copy[key] = value
         # check dimension consistency
@@ -245,7 +272,6 @@ class OnlineFlow(mdp.Flow):
         self._train_arg_keys_list, self._train_args_needed_list = self._get_required_train_args_from_flow(flow_copy)
 
     def __delitem__(self, key):
-        # make a copy of list
         flow_copy = list(self.flow)
         del flow_copy[key]
         # check dimension consistency
@@ -335,12 +361,38 @@ class CircularOnlineFlowException(mdp.MDPException):
 
 class CircularOnlineFlow(OnlineFlow):
     """A 'CircularOnlineFlow' is a cyclic sequence of online/non-trainable nodes that are trained and executed
-    together to form a more complex algorithm.  The input and output nodes of the flow can be changed
-    at any time using the functions "set_input_node" and "set_output_node".
+    together to form a more complex algorithm. This type of flow is useful when one needs to train the nodes
+    internally for several iterations using stored inputs before training with the next external input.
 
-    CircularOnlineFlow also supports training without external input data. In this case, the flow iterates
-    using the previously stored output as the input. See the docstring of the train method
-    for more information about different training types.
+    input=x(i) -> [node1, node2, node3] -> output=y'     ----External input training----         (1 iteration)
+
+    input=y'   -> [node1, node2, node3] -> output=y'     ---- Internal
+    ...                                                               input                      (n-1 iterations)
+    input=y'   -> [node1, node2, node3] -> output=y(i)                      training----
+
+    This type of processing is especially useful for implementing control algorithms, reinforcement learning, etc.
+
+    The input and the output nodes of the flow can be changed at any time using the functions "set_input_node"
+    and "set_output_node".
+
+    Examples:
+
+    If input node and output node are set equal to node1 :
+
+     Input -> [node1] -> output
+             /       \
+          [node3] <- [node2]
+
+    If input node = node1 and output node is equal to node2
+
+     Input -> [node1] -> [node2] -> output
+                    \    /
+                   [node3]
+
+    CircularOnlineFlow also supports training nodes while ignoring external input data at all times. In this case,
+    the flow iterates everytime using the previously stored output as the input. Input data can be arbitrary (with same
+    input_dim and dtype) and is only used as a clock signal to trigger flow processing.
+    See the docstring of the train method for more information about different training types.
 
     Crash recovery is optionally available: in case of failure the current
     state of the flow is saved for later inspection.
@@ -421,10 +473,9 @@ class CircularOnlineFlow(OnlineFlow):
     def train(self, data_iterables):
         """Train all trainable-nodes in the flow.
 
-        'data_iterables' is a single iterable (including generator-type iterators if
-        the last node has no multiple training phases) that must return data
-        arrays to train nodes (so the data arrays are the 'x' for the nodes).
-        Note that the data arrays are processed by the nodes
+        'data_iterables' is a single iterable (including generator-type iterators)
+        that must return data arrays to train nodes (so the data arrays are the 'x'
+        for the nodes). Note that the data arrays are processed by the nodes
         which are in front of the node that gets trained,
         so the data dimension must match the input dimension of the first node.
 
--- a/mdp/linear_flows_online.py
+++ b/mdp/linear_flows_online.py
@@ -21,16 +21,21 @@ class OnlineFlow(mdp.Flow):
     training (including supervised training and multiple training phases),
     execution, and inverse execution (if defined) of the whole sequence.
 
-    OnlineFlow sequence can contain
-    (a) only OnlineNodes, or
-    (b) a mix of OnlineNodes and trained or non-trainable Nodes (classic), or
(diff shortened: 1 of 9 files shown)


mdp/linear_flows_online.py
```diff
@@ -0,0 +1,607 @@
...
+                _cache['node#%d' % i] = node.cache
+        return _cache
+
+    def __setitem__(self, key, value):
+        if isinstance(key, slice):
+            [self._check_value_type_is_compatible(item) for item in value]
+        else:
+            self._check_value_type_is_compatible(value)
+
+        # make a copy of list
```
nwilbert: I'd leave out comments like these that just describe what the code is doing (without providing real value).

mdp/linear_flows_online.py
```diff
@@ -0,0 +1,607 @@
...
+
+class CircularOnlineFlowException(mdp.MDPException):
+    """Base class for exceptions in Flow subclasses."""
+    pass
+
+
+class CircularOnlineFlow(OnlineFlow):
+    """A 'CircularOnlineFlow' is a cyclic sequence of online/non-trainable nodes that are trained and executed
+    together to form a more complex algorithm.  The input and output nodes of the flow can be changed
+    at any time using the functions "set_input_node" and "set_output_node".
```
nwilbert: This seems like a finer detail, that could be placed further down.

mdp/test/test_onlinenode.py
```diff
@@ -0,0 +1,110 @@
...
+    flow = BogusOnlineNode() + BogusOnlineNode()
+    assert(isinstance(flow, mdp.OnlineFlow))
+    inp = mdp.numx.zeros([2,5])
+    flow.train(inp)
+    out = flow.execute(inp)
+    assert_array_equal(mdp.numx.ones(inp.shape)*2, out)
+    for node in flow:
+        assert(node.is_training())
+
+    flow = BogusOnlineNode() + BogusNodeTrainable()
```
nwilbert: Shouldn't this be a separate test?

mdp/test/test_online_flows_hinet.py
```diff
@@ -0,0 +1,305 @@
...
+    flow.append(newnode)
+    assert_equal(len(flow), length+1)
+    length = len(flow)
+    try:
+        newnode = BogusNode(input_dim=11)
+        flow.append(newnode)
+        raise Exception('flow.append appended inconsistent node')
+    except ValueError:
+        assert_equal(len(flow), length)
+    # extend
```
nwilbert: I'd split this into separate tests.

mdp/linear_flows_online.py
```diff
@@ -0,0 +1,607 @@
...
+    first node and is successively processed by the subsequent nodes along
+    the sequence.
+
+    Using an online flow as opposed to handling manually a set of nodes has a
+    clear advantage: The general online flow implementation automatizes the
+    training (including supervised training and multiple training phases),
+    execution, and inverse execution (if defined) of the whole sequence.
+
+    OnlineFlow sequence can contain trained or non-trainable Nodes and optionally
+    a terminal trainable Node.
```
nwilbert: I think we generally have to be a little careful with the term "training" and "trainable". On the one hand there are "online" nodes that "learn" in their executable state (i.e., with node.execute). On the other hand there are "trainable" nodes, that have standard training phases.
Here's my current understanding (please correct me if this is off): What OnlineFlow really does is to enable specific additional data (supervision signals) to be passed to each node in the flow during the training of the last node in the flow (that is allowed to be trainable).

mdp/linear_flows_online.py
```diff
@@ -0,0 +1,607 @@
...
+       number of nodes or a single numpy array. OnlineFlow requires only one
+       input dataiterable as each node is trained simultaneously.
+
+    c) Additional train args (supervised labels etc) are passed to each node through the
+       node specific dataiterable. OnlineFlow requires the dataiterable to return a list
+       that contains tuples of args for each node: [x, (node0 args), (node1 args), ...]. See
+       train docstring.
+
+    d) OnlineFlow also has the cache attribute that retrieves any caches variables
+       stored in the OnlineNodes.
```
nwilbert: What does this cache exactly do? Is it coupled to the concept of online learning, or is here because it was useful for some specific case?

mdp/linear_flows_online.py
varunrajk: Sure.

varunrajk:
I seemed to have messed up the branch and pull request while working on the edits. I should hopefully be able to restore it back.

mdp/test/test_online_flows_hinet.py
varunrajk: I simply added this test from 'test_flows.py' with modifications to include online flow. In any case, here each list method test uses common values. So, splitting them would require rewriting all the tests.

mdp/linear_flows_online.py
varunrajk: From reverse engineering mdp :) Here is my understanding between a few terms (correct me if I am wrong):
trainable - This defines whether a node is trainable or not. That is, is_trainable() returns True.
training - This is actually a state of a trainable node. If training completes is_training() returns False. 

A "trainable" node implies that is_trainable() = True. It does not indicate the current state of the node, that is, the training may have been finished or not. is_training() = True or False.
A "non-trainable" node implies that is_trainable() returns False and so does is_training().
A "trained" node implies that is_trainable() returns True, but is_training() returns False. 

So, an online flow could be a sequence of 
(a) OnlineNodes, or
(b) a mix of OnlineNodes and "trained" or "non-trainable" nodes, or
(c) a mix of OnlineNodes/trained/non-trainable nodes and a terminal "trainable" node. 

An OnlineFlow trains all the nodes simultaneously for the given input sample x. 
So, node2 is training while node1 is still training (node1.is_training() = True)

Whereas a standard Flow trains nodes sequentially, that is, the second node is only trained after the first node's training has finished  (node1.is_training() = False)

I have tried to make sure that online nodes only "learn" when their train method is called. execute method does not change the state of the algorithm only train method does.

varunrajk: Please let me know if this makes sense. Then I can improve the doc string of the Online flow to disambiguate some of the terms used.

mdp/linear_flows_online.py
varunrajk: Yes, cache can store some intermediate variables that are not an output via an execute call. These can be useful for debugging/plotting etc. User can choose the variables to monitor during training.

mdp/nodes/stats_nodes_online.py
```diff
@@ -0,0 +1,93 @@
...
+     Also supports exponentially weighted moving average when
+     the parameter avg_n is set.
+
+    This is an online learnable node (OnlineNode)
+
+    **Internal variables of interest (stored in cache)**
+
+      ``self.avg``
+          The current average of the input data
+    """
```
pberkes: You need to document what the execution does.

mdp/nodes/stats_nodes_online.py
```diff
@@ -0,0 +1,93 @@
+import mdp
+
+
+class MovingAvgNode(mdp.PreserveDimOnlineNode):
```
pberkes: If I understand correctly, this is not really a "moving average" in the usual sense (e.g. https://en.wikipedia.org/wiki/Moving_average). It's more of an online average calculation, so I would suggest to call it "OnlineMeanNode", "OnlineAvgNode", or similar

mdp/nodes/stats_nodes_online.py
```diff
@@ -0,0 +1,93 @@
...
+            typ = 'dtype=None'
+        else:
+            typ = "dtype='%s'" % self.dtype.name
+        numx_rng = "numx_rng=%s" % str(self.numx_rng)
+        avg_n = "avg_n=%s" % self.avg_n
+        args = ', '.join((inp, out, typ, numx_rng, avg_n))
+        return name + '(' + args + ')'
+
+
+class MovingTimeDiffNode(mdp.PreserveDimOnlineNode):
```
pberkes: This is also not "moving" in the sense of "computed in a moving window". Maybe again "OnlineTimeDiffNode"?

pberkes:
There's a conflict that stops Travis CI from running.

pberkes:
Thanks fo the PR @varunrajk ! I added my review

mdp/linear_flows_online.py
```diff
@@ -0,0 +1,658 @@
...
+    """
+
+    def __init__(self, flow, crash_recovery=False, verbose=False):
+        super(OnlineFlow, self).__init__(flow, crash_recovery, verbose)
+        # check if the list of nodes is compatible.
+        self._check_compatibilitiy(flow)
+        # collect cache from individual nodes.
+        self._cache = self._get_cache_from_flow(flow)
+        # collect train_args for each node
+        self._train_arg_keys_list, self._train_args_needed_list = self._get_required_train_args_from_flow(flow)
```
varunrajk: Yes it is needed and is checked for OnlineFlows too. OnlineFlow does not overwrite  "_check_nodes_consistency" method from its parent class.

mdp/linear_flows_online.py
```diff
@@ -0,0 +1,658 @@
...
+    def _check_value_type_is_compatible(self, value):
+        # onlinenodes, trained and non-trainable nodes are compatible
+        if not isinstance(value, mdp.Node):
+            raise TypeError("flow item must be a Node instance and not %s" % type(value))
+        elif isinstance(value, mdp.OnlineNode):
+            pass
+        else:
+            # classic mdp Node
+            if value.is_training():
+                raise TypeError("flow item must either be an OnlineNode instance, a trained or a non-trainable Node.")
```
varunrajk: I think there seems to be a confusion with the compatible node sequences for an OnlineFlow. I improved the docstrings and also renamed a method to make this clear. To answer your question let me first discuss the compatible node sequences:

OnlineFlow node sequence can contain
 (a) only OnlineNodes
        (Eg. [MovingAvgNode(), IncSFANode()],
 or
(b) a mix of OnlineNodes and trained/non-trainable Nodes
        (eg. [a fully trained PCANode, IncSFANode()] or [QuadraticExpansionNode(), IncSFANode()],
or
 (c) a mix of OnlineNodes/trained/non-trainable Nodes and a terminal trainable Node (but not an OnlineNode) whose training hasn't finished
        (eg. [IncSFANode(), QuadraticExpansionNode(), a partially or untrained SFANode]).

(a) is fairly obvious. (b) allows inclusion of non-trainable or pre-trained nodes. For example, a pretrained three layered hierarchical SFA node and a top incremental SFA node. (c) allows an uncommon combination of online preprocessing for training a batch node. 

The confusion could also stem from the method names I chose before:
"_check_value_type_is_compatible" (an interim compatibility check)
and 
"_check_compatibilitiy" (this is the actual one)

 I changed these to 
"_check_value_type_is_online_or_nontrainable_node" (renamed)
and 
"_check_compatibilitiy" (same as before).

I hope this clarifies your doubt. Thanks.

mdp/linear_flows_online.py
```diff
@@ -0,0 +1,658 @@
...
+
+    def _check_compatibilitiy(self, flow):
```
varunrajk: Thanks.

mdp/test/__init__.py
```diff
@@ -27,7 +27,8 @@ def test(filename=None, seed=None, options='', mod_loc=None):
     # add --assert=reiterp option to work around permissions problem
     # with __pycache__ directory when MDP is installed on a normal
     # user non-writable directory
-    options = "--assert=reinterp "+options
+    # options = "--assert=reinterp "+options
+    options = "-s "+options
```
otizonaizit: I think here the problem is that you did not merge master on your branch recently? The "assert=reinterp" option to py.test has been removed with the last commit in master: d44efa5 . As that was a revert commit, it creates a conflict. Damn, next time I should simply revert manually and create a new commit...

mdp/nodes/stats_nodes_online.py
varunrajk: There are two parts to this comment:

First on the definition of the moving average:
MovingAvgNode supports two averages: (a) A cumulative moving average according to the wiki page you have pointed out, which is also an online average calculation. (b) An exponentially weighted moving average. All moving averages more or less follow an update equation similar to:
avg_t = alpha * x_t + (1-alpha) * avg_t-1. For cumulative moving avg: alpha = 1/t and for exponential it is a constant. So I believe in this regard the usage of moving average as its name is justified. 

 However, after taking a second look, I figured out that this may not be a right name for the node after all. Since the node does not return an average but returns the centered data. So, I aptly changed the name of the node to OnlineCenteringNode and improved the docstring. The node's main purpose is to center the data via an execute call by updating an average via the train call. 

Thanks for pointing this out anyway :)

mdp/nodes/stats_nodes_online.py
varunrajk: Yes, I completely agree with you on this. Changed it to OnlineTimeDiffNode and improved the doc string.

pberkes:
Hi @varunrajk . I'm happy with the revision. The only thing I'm concerned about it is commit https://github.com/mdp-toolkit/mdp-toolkit/pull/20/commits/a4910b987a4956c34c965ad6a44caab9811cc949 : Was it a bug as the commit title suggests? Then please add a test reproducing the bug.

varunrajk:
Hi @pberkes, Thanks for the approval. I added the necessary test function to check for the bug.

mdp/nodes/stats_nodes_online.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(self, avg_n=None, input_dim=None, output_dim=None, dtype=None, numx_rng=None):
+        """
+        avg_n - If set to None (default), the node updates a simple moving average. If set to a positive integer,
+        the node updates an exponentially weighted moving average.
+        """
+        super(OnlineCenteringNode, self).__init__(input_dim=input_dim, output_dim=output_dim, dtype=dtype,
+                                                  numx_rng=numx_rng)
+        self.avg_n = avg_n
+        self.avg = None
+        self._cache = {'avg': None}
```
nwilbert: Maybe I'm missing something, but where is this cache used?

mdp/nodes/stats_nodes_online.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def _check_params(self, x):
+        if self.avg is None:
+            self.avg = mdp.numx.zeros(x.shape[1], dtype=self.dtype)
+
+    def _get_supported_training_types(self):
```
nwilbert: Is this method actually used? I can't remember it.

mdp/signal_node_online.py
```diff
@@ -0,0 +1,376 @@
...
+        If numx_rng is unspecified, it will be set to a random number generator
+        with a random seed.
+        """
+        super(OnlineNode, self).__init__(input_dim, output_dim, dtype)
+        # this var stores the index of the current training iteration
+        self._train_iteration = 0
+        # this cache var dict stores an interemediate result or paramenter values
+        # at the end of each train iteration. OnlineNode subclasses should
+        # initialize the required keys
+        self._cache = dict()
```
nwilbert: What is the advantage of having the cache dict, instead of dedicated variables in subclasses?
Maybe I missed it, but I don't see any general functionality that is enabled by this?

mdp/linear_flows_online.py
```diff
@@ -0,0 +1,691 @@
...
+       number of nodes or a single numpy array. OnlineFlow requires only one
+       input dataiterable as each node is trained simultaneously.
+
+    c) Additional train args (supervised labels etc) are passed to each node through the
+       node specific dataiterable. OnlineFlow requires the dataiterable to return a list
+       that contains tuples of args for each node: [x, (node0 args), (node1 args), ...]. See
+       train docstring.
+
+    d) OnlineFlow also has the cache attribute that retrieves any temporary caches variables
+       stored in the OnlineNodes.
```
nwilbert: As you can see from the comments I created an then deleted I'm a bit confused about the benefit of the cache.
Can you describe what it is supposed to do?
Is it deeply connected to the concept of online nodes/flows? Otherwise it might be better to separate it out (possibly as an extension or by some other means).

varunrajk: Hi @nwilbert. When I started coding the OnlineNode, I decided to keep a dict to store a few  variables that could be useful for debugging/plotting. However as the rest of the code matured around the OnlineNode, I rarely use(d) it in the way that I had initially intended. And now after giving a second thorough look at it, I seem to agree with you. It is not actually necessary and perhaps adds redundancy to the code. In any case, if in the future if I find it really necessary, then yes, an extension could be an option. Thanks for pointing out this "vestigial" feature :). I will get rid of it and push the changes.

PR #23, opened 2017-02-18
Code review

varunrajk:
Hi, 

Here is my updated second PR.
Commit messages have all the information of the new changes.

### DIFF
--- a/mdp/hinet/flownode.py
+++ b/mdp/hinet/flownode.py
@@ -140,9 +140,6 @@ class FlowNode(mdp.Node):
         """Return a training sequence containing all training phases."""
         
         def get_train_function(_i_node, _node):
-            # This internal function is needed to channel the data through
-            # the nodes in front of the current nodes.
-            # using nested scopes here instead of default args, see pep-0227
             def _train(x, *args, **kwargs):
                 if _i_node > 0:
                     _node.train(self._flow.execute(x, nodenr=_i_node-1),
--- a/mdp/hinet/flownode.py
+++ b/mdp/hinet/flownode.py
@@ -54,8 +54,10 @@ class FlowNode(mdp.Node):
                               - self._pretrained_phase[i_node])
         if train_len:
             self._is_trainable = True
+            self._training = True
         else:
             self._is_trainable = False
+            self._training = False
         # remaining standard node initialisation
         super(FlowNode, self).__init__(input_dim=input_dim,
                                        output_dim=output_dim, dtype=dtype)
@@ -142,7 +144,7 @@ class FlowNode(mdp.Node):
             # the nodes in front of the current nodes.
             # using nested scopes here instead of default args, see pep-0227
             def _train(x, *args, **kwargs):
-                if i_node > 0:
+                if _i_node > 0:
                     _node.train(self._flow.execute(x, nodenr=_i_node-1),
                                 *args, **kwargs)
                 else:
@@ -219,3 +221,4 @@ class FlowNode(mdp.Node):
 
     def __iter__(self):
         return self._flow.__iter__()
+
--- a/mdp/hinet/layer.py
+++ b/mdp/hinet/layer.py
@@ -51,16 +51,9 @@ class Layer(mdp.Node):
             self.node_input_dims[index] = node.input_dim
         output_dim = self._get_output_dim_from_nodes()
 
-        # store which nodes are pretrained up to what phase
-        _pretrained_phase = [node.get_current_train_phase()
-                             for node in nodes]
-        # check if all the nodes are already fully trained
-        train_len = 0
-        for i_node, node in enumerate(nodes):
-            if node.is_trainable():
-                train_len += (len(node._get_train_seq())
-                              - _pretrained_phase[i_node])
-        if train_len:
+        # set layer state
+        nodes_is_training = [node.is_training() for node in nodes]
+        if mdp.numx.any(nodes_is_training):
             self._is_trainable = True
             self._training = True
         else:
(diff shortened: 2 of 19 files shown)


varunrajk:
Travis throws "TypeError: norm() got an unexpected keyword argument 'axis'". But numpy/scipy linalg norm accept axis as a keyword. Strange. Moreover, all the tests are passed on my machine.

mdp/hinet/layer.py
```diff
@@ -49,6 +50,23 @@ def __init__(self, nodes, dtype=None):
...
+
+        # store which nodes are pretrained up to what phase
+        _pretrained_phase = [node.get_current_train_phase()
+                             for node in nodes]
+        # check if all the nodes are already fully trained
+        train_len = 0
+        for i_node, node in enumerate(nodes):
+            if node.is_trainable():
+                train_len += (len(node._get_train_seq())
+                              - _pretrained_phase[i_node])
```
nwilbert: You can inline the get_current_train_phase call here, since you don't really need the _pretrained_phase list. Or am I missing something?

mdp/hinet/layer.py
```diff
@@ -116,7 +134,7 @@ def _get_supported_dtypes(self):
         return list(types)
 
     def is_trainable(self):
-        return any(node.is_trainable() for node in self.nodes)
+        return self._is_trainable
```
nwilbert: Did you run into efficiency proplems with the old code? I'm wondering at what point this makes a measurable difference.

mdp/hinet/layer.py
```diff
@@ -283,6 +314,23 @@ def __init__(self, nodes, dtype=None):
...
+        for i_node, node in enumerate(nodes):
+            if node.is_trainable():
+                train_len += (len(node._get_train_seq())
+                              - _pretrained_phase[i_node])
+        if train_len:
+            self._is_trainable = True
+            self._training = True
+        else:
+            self._is_trainable = False
+            self._training = False
```
nwilbert: Since this code seems to be identical to that from the normal layer, wouldn't it be better to extract it into a common method or something?

mdp/hinet/flownode.py
```diff
@@ -142,7 +144,7 @@ def get_train_function(_i_node, _node):
             # the nodes in front of the current nodes.
             # using nested scopes here instead of default args, see pep-0227
             def _train(x, *args, **kwargs):
-                if i_node > 0:
+                if _i_node > 0:
```
nwilbert: Seems like this was indeed a bug, though it would only affect the efficiency of the code. I wonder if we need better testing for this method.

mdp/hinet/flownode.py
```diff
@@ -142,7 +144,7 @@ def get_train_function(_i_node, _node):
             # the nodes in front of the current nodes.
             # using nested scopes here instead of default args, see pep-0227
```
nwilbert: I think we can delete this line, I think I wrote it when nested scope in Python was something new and fancy ;-) (or at least seemed like that to me...)

mdp/hinet/layer.py
varunrajk: Yes, you are right. pre trained phases are not required. I just wanted to know if the node is still training or not. But I can just check if the node's is_training flag is set. Perhaps, I moved this code from the flow when I was not clear about the distinction between trained/trainable/non-trainable nodes. Will fix this. Thanks.

mdp/hinet/layer.py
varunrajk: Well it is mostly for consistency. When creating a layer, we need to check not just whether the individual nodes are trainable but if there are any remaining training phases. For example, a layer made up of trained SFAs would still return True when is_trainable is called. But in principle the layer is a non-trainable node.

mdp/hinet/layer.py
varunrajk: Sorry, my bad. Yes, this is not needed.

mdp/hinet/flownode.py
varunrajk: Sure :)

mdp/hinet/layer.py
varunrajk: I will add a test function to check this.

mdp/hinet/layer.py
varunrajk: Sorry deleted my previous comment. The reason this part of the code is repeated is because SameInputLayer does not call its super where the flags are set. It instead calls Node's constructor.

mdp/hinet/layer.py
nwilbert: Ok, so it is more about consistency with FlowNode than an actual bug?

mdp/hinet/layer.py
nwilbert: Oh god, yeah, I used an MRO-hack in SameInputLayer. I guess at the time I thought that this was smart, and didn't realize that I was violating the Liskov substitution principle. Now I think that SameInputLayer shouldn't really derive from Layer.

mdp/hinet/layer_online.py
```diff
@@ -0,0 +1,149 @@
+"""
+Module for OnlineLayers.
```
nwilbert: I wonder if it would be better to put all the online stuff into a new subpackage (like hinet and parallel)?
This is just a thought, I haven't really thought it through yet.

varunrajk: Yes, I too had this as a thought initially. In fact in my forked mdp copy, you can find among my earlier commits that I had followed this style of separate online folder. But after discussing at length here at the INI group meetings, we concluded that this was not such a great idea :)

mdp/hinet/layer.py
varunrajk: Thanks for pointing out the Liskov subs. princ. I'll sort this out in the next PR.

mdp/hinet/layer_online.py
nwilbert: Ok, we can always reconsider this later.

mdp/hinet/layer.py
nwilbert: No problem. It would be nice if this is fixed, but since this is an old issue this isn't urgent.

PR #25, opened 2017-03-15
Code review

varunrajk:
Hi Everyone, 

Here's my third PR. 

Added quite a few nodes including openai gym interface and fast online plotting nodes.

Varun.


mdp/hinet/switchboard.py
```diff
@@ -486,21 +479,16 @@ def __init__(self, in_channels_xy, field_channels_xy,
...
+                        connections[first_out_con:first_out_con + in_channel_dim] = list(
+                            range(first_in_con, first_in_con + in_channel_dim))
                         first_out_con += in_channel_dim
-        super(DoubleRect2dSwitchboard, self).__init__(
-                                input_dim=in_channel_dim *
-                                    in_channels_xy[0] * in_channels_xy[1],
-                                connections=connections,
-                                out_channel_dim=out_channel_dim,
-                                in_channel_dim=in_channel_dim)
+        super(DoubleRect2dSwitchboard, self).__init__(input_dim=in_channel_dim * in_channels_xy[0] * in_channels_xy[1],
```
nwilbert: I'm wondering about the max line length. I think so far we followed the PEP8 convention of 79 characters max. I'm open to relaxing this to e.g., 99 chars (as suggested by PEP8) or possibly higher (I've been in some heated discussions on this topic and have grown more flexible on this). But I would be interested in what the other maintainers think, and we should reflect this in the MDP development guide.

mdp/hinet/switchboard.py
```diff
@@ -688,3 +669,99 @@ def index_to_image(self, index):
...
+class RandomChannelSwitchboardException(mdp.NodeException):
+    """Exception for routing problems in the Switchboard class."""
+    pass
+
+
+class RandomChannelSwitchboard(ChannelSwitchboard):
+    """
+    Randomly sampled 2D Switchboard.
+    Generates input-output connections for hierarchical networks with randomly sampled receptive fields.
+    New connections are set on every execute call
```
nwilbert: Could you add a brief about possible use cases for this node?

mdp/hinet/switchboard.py
```diff
@@ -688,3 +669,99 @@ def index_to_image(self, index):
...
+            self._lut_conn[y_start_chan, x_start_chan] = connections
+
+        connections = self._new_connections()
+        super(RandomChannelSwitchboard, self).__init__(input_dim=(in_channel_dim * in_channels_xy[0] *
+                                                                  in_channels_xy[1]),
+                                                       connections=connections,
+                                                       out_channel_dim=out_channel_dim,
+                                                       in_channel_dim=in_channel_dim)
+
+    def _new_connections(self):
```
nwilbert: I'd pick a more specific name, like "random_connections".
I'd either make this method purely functional (providing the required data as arguments) or fully go the object oriented route (by overwriting the connections in the method and calling it something like "update_random_connections").

mdp/nodes/basis_function_nodes.py
```diff
@@ -0,0 +1,466 @@
+import mdp
+from mdp.utils import mult
+import itertools
+from misc_nodes import GridProcessingNode
+
+
+class BasisFunctionNode(mdp.Node):
+    """
+    BasisFunctionNode is a non-trainable node that
+    provides a set of non-linear basis functions,
```
nwilbert: I don't really understand what this node does based on this docstring. What does "provides" mean in this context?

mdp/nodes/basis_function_nodes.py
```diff
@@ -0,0 +1,466 @@
...
+
+        self._gp = GridProcessingNode(grid_lims=lims, n_grid_pts=n_grid_pts)
+
+        self._input_dim = len(lims[0])
+
+        self._basis_mode = 'continuous'
+
+        # Basis Fns Init
+
+        if self.basis_name is 'indicator':
```
nwilbert: This method is getting quite long ;-). You could either have subclasses for the different function types, or use something like the strategy pattern (which is probably nicer). If you use the strategy pattern then this might also be a chance to extract the get_fn_responses_img helper method, since I don't think that something like this should be part of a node class.

mdp/nodes/hsfa_nodes.py
```diff
@@ -0,0 +1,540 @@
...
+
+    HSFANode also supports random sampling of data-cuboids via a RandomSwitchboard
+    instead of the fixed RectangularSwitchboard.
+
+    HSFANode also implements most of the 'list' container methods; such as insert,
+    add, delete layers from the network.
+
+    The node also provides a 'show_flow' utility method to visualize the
+    network.
+
```
nwilbert: What is the advantage of this class over the FlowNode approach?
I'm also worried that this class includes too much different functionality (e.g., that it has a show_flow method). It sounds a bit like a god class.
I'd probably rather provide a helper function to create HSFA networks.

mdp/nodes/misc_nodes.py
```diff
@@ -815,3 +826,603 @@ def _execute(self, x):
...
+                            'img_255_1' - scales uint img data to float values between [0, 1]
+                            'gray' - converts to grayscale images.
+                            'to_2d' - converts data to 2D
+                            'to_dtype' - converts data to the desired data type
+                            'set_shape' - reshapes the data to the given shape. If shape argument is not provided,
+                            then it uses a stored shape buffer, which is either the input_shape or the shape
+                            lost if 'to_2d' is executed.
+
+        trasnform_seq_args - A list of required arguments tuples for each transformation, ordered according to
+                            the transform_seq.
```
nwilbert: This sound to me like a candidate for the strategy pattern, or different subclasses (similar to the situation with the BasisFunctionNode).

mdp/nodes/misc_nodes.py
```diff
@@ -815,3 +826,603 @@ def _execute(self, x):
...
+
+    # utility methods
+
+    def get_transform_fn_names(self):
+        """Returns the available transformation function names"""
+        fns = self._get_transform_fns()
+        return fns.keys()
+
+
+# TODO: Could use or add to the graph package of MDP?
```
nwilbert: Do you think it might be better to use networkx as a base instead? While networkx is far from perfect, it might have far better performance.

mdp/nodes/misc_nodes.py
```diff
@@ -815,3 +826,603 @@ def _execute(self, x):
...
+                    'nuc'  nuclear norm                  --
+                    inf    max(sum(abs(x), axis=1))      max(abs(x))
+                    -inf   min(sum(abs(x), axis=1))      min(abs(x))
+                    0      --                            sum(x != 0)
+                    1      max(sum(abs(x), axis=0))      as below
+                    -1     min(sum(abs(x), axis=0))      as below
+                    2      2-norm (largest sing. value)  as below
+                    -2     smallest singular value       as below
+                    other  --                            sum(abs(x)**ord)**(1./ord)
+                    =====  ============================  ==========================
```
nwilbert: Maybe one could use an enum or something like that (we would have to use enum34 for backwards compatibility). This might also be another chance for the strategy pattern (e.g., with a factory function for l-norm functions)

mdp/nodes/openai_gym_nodes.py
```diff
@@ -0,0 +1,293 @@
...
+        self.env = gym.make(self.env_name)
+
+        self.render = render
+        self.auto_reset = auto_reset
+
+        # set a shared numx_rng
+        self.numx_rng = numx_rng
+
+        # get observation dims and shape
+        if isinstance(self.env.observation_space, spaces.discrete.Discrete):
```
nwilbert: It would be nice to use polymorphism instead of explicit type checks. But I don't yet understand the code deep enough to tell if this would be feasible.

mdp/nodes/hsfa_nodes.py
varunrajk: Well there are quite a few advantages: 
(a) A standardized node to implement HSFA that everyone can use. I have often heard from people complain that they get different results or sometimes terrible results. This is usually because of a few blunders in their code. 

(b) HSFANode simplifies the creation of a hsfa net without having to worry about switcboards/interconnections etc. This is especially for beginners who simply want to use HSFANode without needing to deeply understand how mdp works. The node also allows to autoset a few params by using -1.

(c) HSFANode also supports using random sampled receptive fields during training via the RandomSwitchboard. This can speed up training time on high-resolution images. 

(d) HSFANode also supports container methods like adding a layer or setting a layer with checks. This is not possible straightway with flownodes. Flownode does not include additional checks that are required to define a proper HSFANode. 

I have added a show flow just to further customize the visualization for hsfanode.

mdp/nodes/openai_gym_nodes.py
varunrajk: I am not sure how we can use polymorphism here. Any suggestions are welcome :)

mdp/nodes/misc_nodes.py
varunrajk: Doesn't numpy/scipy norm take care of this? I am not explicitly calculating different types of norms in the code, instead I am just passing the nbr_dist_fn as an argument to numpy/scipy norm. Please tell me if I misunderstood your comment.

mdp/nodes/misc_nodes.py
varunrajk: Thanks for pointing this out. I will update the todo and will take a look into it.

mdp/hinet/switchboard.py
varunrajk: Sure.

mdp/hinet/switchboard.py
varunrajk: I use pycharm and organize my code according to the pep8 warnings it throws. Maybe this information is worth taking into consideration?

nwilbert: I think PyCharm uses a default value of 120 for the "right margin". The line length code inspection might also be disabled by default.
I'd be interested to hear what Tiziano and the others think about this. I'm fine with both options (strictly following PEP8 or raising the length limit).

mdp/nodes/misc_nodes.py
nwilbert: Ah, I see, I overlooked that this is simply forwarded to numpy. I see that you already described this in the docstring, sorry.

mdp/nodes/misc_nodes.py
```diff
@@ -815,3 +826,603 @@ def _execute(self, x):
...
+        n_grid_pts - number of grid points within each dimension. This provides the spacing between
+                the grid points. Default is set to the number of integers between the bounds.
+
+        output_type - the type of conversion. Accepted values:
+                    'gridx' - default (just returns the input)
+                    'graphx' - returns graph coordinates
+                    'graphindx' - returns graph indices of the input.
+
+        nbr_dist_fn - distance function to define the closest neighbors (applied on the graph
+                    coordinates). This is the same as
```
nwilbert: I'd change the wording just a little: "This is passed as the "order" argument to the numx...".

Maybe we can improve the argument name a little, the "fn" ending sounds to me like this should be a callable. "nbr_norm_ord" for example would be closer to the argument name in NumPy.

mdp/hinet/switchboard.py
varunrajk: You are right about the 120 value! Strange that they don't use pep8 value as the default. Thanks for pointing this out. I would vote for pep8 limit of 79, since we are anyway following pep8 convention for the rest.

mdp/nodes/misc_nodes.py
varunrajk: Sounds good. Thanks!

otizonaizit:
@varunrajk , @nwilbert : What's the status of this PR? Are we coming close to merging? What needs to be done?

nwilbert:
@otizonaizit : Sorry, I had kind of forgotten about this :-(. I can have a fresh look at it on the coming weekend.
@varunrajk : I assume that you were quite busy with the new job. Are you still interested in moving this forward? Sorry to keep you waiting.

varunrajk:
Hi, I am really sorry that I could not follow up. I have a deadline mid of this month. After which, I will have time to work on this PR.

varunrajk:
Unfortunately, this PR is something that I cannot take up right now. So, I'm closing it. Hopefully, if there is something useful in here, someone can pull out some code from here.

PR #30, opened 2017-08-31
SFANode update

fabschon:
This Pull Request includes additional functionality of the basic SFANode:

- **SFANode.__init__ arguments**
  - *use_svd:* Switch to use SVD-based solution to the internal eigenvalue problem instead of the default call to `symeig()`. **Default:** False.
  - *always_analyse_data:* If `_stop_training()` fails with the common "singular values" exception, the code will now produce a simple plot that depicts some basic data analysis (see below). If set to True, this will always happen; when set to False it will only trigger the analysis when the exception is thrown. **Default:** False.

- **SFANode._svd_solver()**
  - Alternative method to solving the eigenvalue problem at the heart of SFA. See [1] for details on the implemented algorithm.
  - The SVD based method requires a bit more memory and slightly more time than the default method. Profiling was done with a simple script found [here](https://pastebin.com/jd74rNPc). Timings can be seen below [2]; memory use can be seen [here](http://i.imgur.com/ttqT1l9.png). (Profiling test case used a random 2k x 2k matrix as `self.cov_mtx` and `self.dcov_mtx` and ran ten times through the respective processes.) The script can quickly be extended to test different methods of solving the internal eigenvalue issue.

- **Debug additions**
  - Three utility functions to provide more information in case `_stop_training()` fails with the common "singluar values" exception: `_data_fail_msg()`; `_data_analysis()`; and `_data_analysis_plotter()`.
  - `_data_fail_msg()` merely returns a more verbose error message if the exception is thrown.
  - `_data_analysis()` produces a plot depicting variance and slowness per signal, as well as the eigenvalue spectrum. it uses `_data_analysis_plotter()` as an utility function.
  - `_data_analysis_plotter()` creates the sub plots for the overall plot created in `_data_analysis()`.

**[1]** http://www.geo.tuwien.ac.at/downloads/tm/svd.pdf (section 0.3.2)
**[2]** SVD and symeig timings:
```
$ ./run_profiling.sh 
+ mprof run --python python2 profiling.py
mprof: Sampling memory every 0.1s
running as a Python program...

Using numpy version 1.12.0 (@ /home/fabian/.local/lib/python2.7/site-packages/numpy)
Using mdp version 3.5 (@ /local/work/MDP-3.5/mdp)
Generating garbage..
Storing garbage in memmap..
Garbage ready for use.

Running symeig_test
    Run 1/10.. done. (11.3755309582)
    Run 2/10.. done. (11.335201025)
    Run 3/10.. done. (11.3295040131)
    Run 4/10.. done. (11.4707131386)
    Run 5/10.. done. (11.4013779163)
    Run 6/10.. done. (11.3865709305)
    Run 7/10.. done. (11.4483189583)
    Run 8/10.. done. (11.6087400913)
    Run 9/10.. done. (11.4543762207)
    Run 10/10.. done. (11.4626760483)
    Avg. time (excluding first run): 11.4330531491
    Min. time (excluding first run): 11.3295040131

Running svd_test
    Run 1/10.. done. (11.5337908268)
    Run 2/10.. done. (14.9500858784)
    Run 3/10.. done. (17.1977508068)
    Run 4/10.. done. (11.8235440254)
    Run 5/10.. done. (12.585725069)
    Run 6/10.. done. (11.9705069065)
    Run 7/10.. done. (13.4548399448)
    Run 8/10.. done. (17.0780851841)
    Run 9/10.. done. (13.7380120754)
    Run 10/10.. done. (12.5583138466)
    Avg. time (excluding first run): 13.9285404152
    Min. time (excluding first run): 11.8235440254
All done.
+ mprof plot
Using last profile data.
```

### DIFF
--- a/mdp/hinet/layer.py
+++ b/mdp/hinet/layer.py
@@ -261,6 +261,12 @@ class CloneLayer(Layer):
         super(CloneLayer, self).__init__((node,) * n_nodes, dtype=dtype)
         self.node = node  # attribute for convenience
 
+    def _execute(self, x, *args, **kwargs):                                               ### VARUN's fix
+        n_samples = x.shape[0]                                                            ### VARUN's fix
+        x = x.reshape(n_samples * x.shape[1] / self.node.input_dim, self.node.input_dim)  ### VARUN's fix
+        y = self.node.execute(x)                                                          ### VARUN's fix
+        return y.reshape(n_samples, self.output_dim)                                      ### VARUN's fix
+
     def _stop_training(self, *args, **kwargs):
         """Stop training of the internal node."""
         if self.node.is_training():
--- a/mdp/nodes/sfa_nodes.py
+++ b/mdp/nodes/sfa_nodes.py
@@ -2,6 +2,8 @@ from builtins import str
 from builtins import range
 __docformat__ = "restructuredtext en"
 
+import numpy as np ############################################################ SVD
+
 import mdp
 from mdp import numx, Node, NodeException, TrainingException
 from mdp.utils import (mult, pinv, CovarianceMatrix, QuadraticForm,
@@ -145,6 +147,10 @@ class SFANode(Node):
 
         rng = self._set_range()
 
+        """
+        ####################################################################### OLD
+        print '[eig]'
+
         #### solve the generalized eigenvalue problem
         # the eigenvalues are already ordered in ascending order
         try:
@@ -161,6 +167,25 @@ class SFANode(Node):
         except SymeigException as exception:
             errstr = str(exception)+"\n Covariance matrices may be singular."
             raise NodeException(errstr)
+        ####################################################################### OLD
+        """
+        ####################################################################### NEW
+        print '[svd]'
+
+        U, s, _ = np.linalg.svd(self.cov_mtx)
+        X1 = np.dot(U, np.diag(1.0 / s ** 0.5))
+        X2, _, _ = np.linalg.svd(np.dot(X1.T, np.dot(self.dcov_mtx, X1)))
+        E = np.dot(X1, X2)
+        e = np.dot(E.T, np.dot(self.dcov_mtx, E)).diagonal()
+
+        e = e[::-1]      # SVD delivers the eigenvalues sorted in reverse (compared to symeig).
+        E = E.T[::-1].T  # -> We reverse the array/matrix storing the eigenvalues/vectors respectively.
+
+        if rng is None:
+            self.d, self.sf = e, E
+        else:
+            self.d, self.sf = e[rng[0] - 1:rng[1]], E[:, rng[0] - 1:rng[1]]
+        ####################################################################### NEW
 
         if not debug:
             # delete covariance matrix if no exception occurred
--- a/mdp/nodes/sfa_nodes.py
+++ b/mdp/nodes/sfa_nodes.py
@@ -1,13 +1,18 @@
-from builtins import str
+
+#--------------------------------------------------------------------[ header ]
+
 from builtins import range
 __docformat__ = "restructuredtext en"
 
-import numpy as np ############################################################ SVD
+import numpy as np
+import matplotlib.pyplot as plt
 
 import mdp
 from mdp import numx, Node, NodeException, TrainingException
 from mdp.utils import (mult, pinv, CovarianceMatrix, QuadraticForm,
                        symeig, SymeigException)
+
+#-------------------------------------------------------------------[ SFANode ]
 
 class SFANode(Node):
     """Extract the slowly varying components from the input data.
@@ -73,10 +78,15 @@ class SFANode(Node):
     """
 
     def __init__(self, input_dim=None, output_dim=None, dtype=None,
-                 include_last_sample=True):
+                 include_last_sample=True, use_svd=False, data_info='never'):
         """
-        For the ``include_last_sample`` switch have a look at the
-        SFANode class docstring.
+        - include_last_sample: see SFANode docstring or API doc [1].
+        - use_svd:   set to True if training throws SymeigException.
+        - data_info: produce additional information about the training data
+                     'never', 'always', or on 'fail' (only if training fails
+                     with SymeigException).
+
+        [1] http://mdp-toolkit.sourceforge.net/api/mdp.nodes.SFANode-class.html
          """
         super(SFANode, self).__init__(input_dim, output_dim, dtype)
         self._include_last_sample = include_last_sample
@@ -87,8 +97,10 @@ class SFANode(Node):
         # one for the derivatives
         self._dcov_mtx = CovarianceMatrix(dtype)
 
-        # set routine for eigenproblem
-        self._symeig = symeig
+        # set parameters for solving the eigenproblem
+        self._symeig   = symeig
+        self._use_svd  = use_svd
+        self._dat_info = data_info
 
         # SFA eigenvalues and eigenvectors, will be set after training
         self.d = None
@@ -119,7 +131,7 @@ class SFANode(Node):
         if  s < 2:
             raise TrainingException('Need at least 2 time samples to '
                                     'compute time derivative (%d given)'%s)
-        
+
     def _train(self, x, include_last_sample=None):
         """
         For the ``include_last_sample`` switch have a look at the
@@ -134,43 +146,81 @@ class SFANode(Node):
         self._cov_mtx.update(x[:last_sample_index, :])
         self._dcov_mtx.update(self.time_derivative(x))
 
-    def _stop_training(self, debug=False):
-        ##### request the covariance matrices and clean up
-        self.cov_mtx, self.avg, self.tlen = self._cov_mtx.fix()
-        del self._cov_mtx
-        # do not center around the mean:
-        # we want the second moment matrix (centered about 0) and
-        # not the second central moment matrix (centered about the mean), i.e.
-        # the covariance matrix
-        self.dcov_mtx, self.davg, self.dtlen = self._dcov_mtx.fix(center=False)
-        del self._dcov_mtx
-
-        rng = self._set_range()
-
+    def _data_plot_(self, x, sub_pos, title='', arg_sort=False, style='dots'):
         """
-        ####################################################################### OLD
-        print '[eig]'
-
-        #### solve the generalized eigenvalue problem
-        # the eigenvalues are already ordered in ascending order
-        try:
-            self.d, self.sf = self._symeig(self.dcov_mtx, self.cov_mtx,
-                                     range=rng, overwrite=(not debug))
-            d = self.d
-            # check that we get only *positive* eigenvalues
-            if d.min() < 0:
-                err_msg = ("Got negative eigenvalues: %s."
-                           " You may either set output_dim to be smaller,"
-                           " or prepend the SFANode with a PCANode(reduce=True)"
-                           " or PCANode(svd=True)"% str(d))
-                raise NodeException(err_msg)
-        except SymeigException as exception:
-            errstr = str(exception)+"\n Covariance matrices may be singular."
-            raise NodeException(errstr)
-        ####################################################################### OLD
+        Helper function for _provide_data_help_. This plots the various subplots
+        produced by _provide_data_help_.
         """
-        ####################################################################### NEW
-        print '[svd]'
+        # plot and axis setup
+        ax = plt.subplot( sub_pos[0], sub_pos[1], sub_pos[2] )
+        if title is not '': ax.set_title( title )
+        x_rng    = np.arange( len(x) )
+        x_labels = x_rng
+        # sort data by dimension (default) or value (arg_sort)
+        if arg_sort:
+            sort = np.argsort( x )
+            x_labels = x_labels[sort]
+            x = x[sort]
+        # plot
+        if style is 'bars':
+            plt.bar( x_rng, x, align='center' )
+        elif style is 'dots':
+            plt.plot( x, 'o' )
+        # ticks and tick labels
+        if len(x) <= 50: plt.xticks( x_rng )
+        ax.set_xticklabels( x_labels )
+        ax.set_xlim( (-1,len(x_rng)) )
+
+    def _provide_data_help(self, full=False ):
+        """
+        This function provides some basic data analysis of the used training
+        data plus more a verbose error message if training the SFANode fails.
+
+        - full: If False, additional information on the training data is given.
+                If True, additional reasoning is given why this might be needed.
+        """
+
+        # plot some basic data analysis of the (presumably failed) training data
+        print '\t..analysing training data..'
+        plt.clf()
+        # plot training data variance per dimension
+        self._data_plot_(np.diagonal(self.cov_mtx), (3, 2, 1), 'Variance per dimension')
+        self._data_plot_(np.diagonal(self.cov_mtx), (3, 2, 2), 'Dimensions by Variance', arg_sort=True)
+        # plot training data slowness/delta values per dimension
+        self._data_plot_(np.diagonal(self.dcov_mtx), (3, 2, 3), 'Slowness per dimension')
+        self._data_plot_(np.diagonal(self.dcov_mtx), (3, 2, 4), 'Dimensions by slowness', True)
+        # plot eigenvalue spectrum
+        _, s, _ = np.linalg.svd( self.cov_mtx )
+        self._data_plot_(s, (3, 2, 5), 'Eigenvalue spectrum')
+        # clean up and save plot
+        plt.tight_layout()
+        plt.savefig( 'SFANode data info.png' )
+        print '\t..training data information plotted..'
+
+        # print optional additional information about the SymeigException
+        err = "This usually happens if there are redundancies in the (expanded) training data.\n" + \
+              "There are several ways to deal with this issue:\n\n" + \
+              "   - Use more data.\n\n" + \
+              "   - Add noise to the data. This can be done by slotting an additional NoiseNode\n\n" + \
+              "     in front of a troublesome SFANode. Noise levels do not have to be high.\n\n" + \
+              "   - Run training data through PCA. This can be done by slotting an additional\n\n" + \
+              "     PCA node in front of the troublesome SFANode. Use the PCA node to discard\n\n" + \
+              "     dimensions within your data with lower variance." + \
+              "   - [ RECOMMENDED ] Use SVD by setting the SFANode flag use_svd to True. This\n\n" + \
+              "     solves the internal SFA eigenvalue problem by using two instances of singular\n\n" + \
+              "     value decomposition (SVD) instead of the default behavior of using the\n\n" + \
+              "     systems's symeig() routine to solve a single, generalized eigenvalue problem.\n\n" + \
+              "     Using SVD is slightly more expensive but more robust.\n\n"
+        return err if full else ''
+
+    def _svd_solver(self, rng):
+        """
+        Alternative routine to solve the SFA eigenvalue issue. This can be used
+        in case the normal symeig() call in _stop_training() throws the common
+        SymeigException ('Covariance matrices may be singular').
+        For details on the used algorithm see:
+            http://www.geo.tuwien.ac.at/downloads/tm/svd.pdf (section 0.3.2)
+        """
 
         U, s, _ = np.linalg.svd(self.cov_mtx)
         X1 = np.dot(U, np.diag(1.0 / s ** 0.5))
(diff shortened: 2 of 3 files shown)


Stewori:
Can we call `use_svd` and `_svd_solver` variables differently, such that it doesn't mention svd.
svd is just an implementation detail, important is that it is a solving method that is robust to rank deficit in the covariance matrix.

We could call it `robust_solver`, `rank_deficit_solver` or `rank_deficit__proof_solver` or so.
The other flag could be called `use_robust_solver` or `use_rank_deficit_proof_solver` or so. (Yes that's too long. Better ideas for a catchy yet precise name?)

I'm pushing for an svd independent name, because there might be other nodes where the same kind of functionality can make sense (ica...?). Depending on the node implementation, not always svd might be the right way to achieve this. Still, we should have a consistent flag name for this kind of tuning.

fabschon:
I'm not sure we want to hide an alternative solution behind a more generic sound. If I was to use SFA and stumbled upon this, my first question would be, well, if this is the "robust" solver, why is it not the default? And when SFA fails on data I'd like to know what is being used to address that (mathematically / algorithmically speaking) without having to look into the source code.

Stewori:
> the "robust" solver, why is it not the default?

Maybe it should be ;) The docstring is required to tell the details (i.e. efficiency considerations) I suppose.

> And when SFA fails on data I'd like to know what is being used to address that

The cause for the issue is rank deficit and svd is not the only way to handle it.
Agreed that the information how it is solved (by default) should be there. I only suggest that it should be in the docstring, so the `_svd_solver` slot can be a slot for any solving method that copes with the rank deficit. Also, `_svd_solver` might be misconcepted to be a pointer to a raw svd algorithm like _symeig points to raw scipy.symeig (IIRC).

mdp/nodes/sfa_nodes.py
```diff
@@ -145,22 +208,33 @@ def _stop_training(self, debug=False):
...
+                    err_msg = ("Got negative eigenvalues: %s."
+                               " You may either set output_dim to be smaller,"
+                               " or prepend the SFANode with a PCANode(reduce=True)"
+                               " or PCANode(svd=True)"% str(d))
+                    raise NodeException(err_msg)
+
+            except SymeigException as exception:
+                errstr  = str(exception)+"\n Covariance matrices may be singular."
+                errstr += self._data_fail_msg()  # offer some additional information on this issue
+                self._data_analysis()            # and print some information about the training set
```
pberkes: If you want to do some diagnostic, it needs to be outside this class. You cannot mix the computational part and the reporting part, users need to be able to use SFANode without any side effects, and do their own reporting if they want.

For example, people who want to do the data analysis might do
```
try:
   node.stop_training()
except NodeException:
  sfa_diagnostic_report(node)
```

mdp/nodes/sfa_nodes.py
```diff
@@ -170,6 +244,9 @@ def _stop_training(self, debug=False):
         # store bias
         self._bias = mult(self.avg, self.sf)
 
+        # plot some statistics of the training data (optional)
+        if self._analysis: self._data_analysis()
```
pberkes: Same here: you have a flag to switch it off, but if somebody wants to do this step they can do it by importing your diagnostic function.

mdp/nodes/sfa_nodes.py
```diff
@@ -181,39 +258,108 @@ def _execute(self, x, n=None):
...
+        """
 
-        The delta value of a signal is a measure of its temporal
-        variation, and is defined as the mean of the derivative squared,
-        i.e. delta(x) = mean(dx/dt(t)^2).  delta(x) is zero if
-        x is a constant signal, and increases if the temporal variation
-        of the signal is bigger.
+        U, s, _ = np.linalg.svd(self.cov_mtx)
+        X1 = np.dot(U, np.diag(1.0 / s ** 0.5))
+        X2, _, _ = np.linalg.svd(np.dot(X1.T, np.dot(self.dcov_mtx, X1)))
```
pberkes: Have you checked with Laurenz that this is meaningful? SFA looks for the smallest eigenvalues of this matrix, and SVD clips eigenvalues that are below a certain threshold, so you might in principle discarding the slowest components.

mdp/nodes/sfa_nodes.py
```diff
@@ -181,39 +258,108 @@ def _execute(self, x, n=None):
...
+               "   - Add noise to the data. This can be done by slotting an additional NoiseNode\n" + \
+               "     in front of a troublesome SFANode. Noise levels do not have to be high.\n\n" + \
+               "   - Run training data through PCA. This can be done by slotting an additional\n" + \
+               "     PCA node in front of the troublesome SFANode. Use the PCA node to discard\n" + \
+               "     dimensions within your data with lower variance.\n\n" + \
+               "   - [ RECOMMENDED ] Use SVD by setting the SFANode flag use_svd to True. This\n" + \
+               "     solves the internal SFA eigenvalue problem by using two instances of singular\n" + \
+               "     value decomposition (SVD) instead of the default behavior of using the\n" + \
+               "     systems's symeig() routine to solve a single, generalized eigenvalue problem.\n" + \
+               "     Using SVD is slightly more expensive but more robust.\n"
```
pberkes: Please define this as a constant instead.

mdp/nodes/sfa_nodes.py
```diff
@@ -1,12 +1,25 @@
-from builtins import str
+
+#====================================================================[ header ]
+
 from builtins import range
 __docformat__ = "restructuredtext en"
 
+import numpy as np
```
pberkes: I realize that `numx` is solving a problem that does not exist anymore, but either you remove it completely everywhere (in another PR), or you keep using it instead of `np`. It does not hurt.

mdp/nodes/sfa_nodes.py
```diff
@@ -145,22 +208,33 @@ def _stop_training(self, debug=False):
...
-            if d.min() < 0:
-                err_msg = ("Got negative eigenvalues: %s."
-                           " You may either set output_dim to be smaller,"
-                           " or prepend the SFANode with a PCANode(reduce=True)"
-                           " or PCANode(svd=True)"% str(d))
-                raise NodeException(err_msg)
-        except SymeigException as exception:
-            errstr = str(exception)+"\n Covariance matrices may be singular."
-            raise NodeException(errstr)
+        if not self._use_svd:
```
pberkes: The right solution here (as mentioned by email) is to do something like
```
if self._use_svd:
    self._symeig = sfa_svd_solver
```

I would to that in the `__init__`, and define `sfa_svd_solver` as a function above `SFANode`.

(My preferred solution is to have a `sfa_solver` parameter in the constructor instead of `use_svd`, so that in principle you can have multiple solver. The default `sfa_solver` would be the `sfa_symeig_solver`. That would avoid the proliferation of on/off parameters if one day you add another solver.)

pberkes: Regarding my preferred solution, to avoid misunderstanding: as far as I'm concerned you can keep the flag, I'm not active anymore and do not want to make design decisions for the new masters of mdp. 

For the other changes requested, I really recommend to follow them as they will spare you some headaches later.

Stewori: I find the `self._symeig = sfa_svd_solver` hackish, because `_symeig` sounds like the field was intended to hold a raw symeig solver and just allow to insert a non default one.
The idea with the `sfa_solver` parameter sounds much better. Consider there is an sfa_solver that uses _symeig internally. Than it would hurt if the sfa node would perform `self._symeig = sfa_solver`, wouldn't it?

pberkes: Maybe we chose the wrong name, that is easy to change. But the attribute and its signature were designed exactly for this case: have a hook to use a different solver. It would be easy to rewrite the `svd_solver` so that it has the same signature:
```
svd_solver(self.dcov_mtx, self.cov_mtx, d=self.d, range=rng, overwrite=True/False)
```
That would make `svd_solver` a standalone function like `symeig`, and we can pass it in the constructor. Nothing else in `_stop_training` needs to change.

Stewori: I guess that's the problem with undocumented parameters - one needs to guess their purpose.

> Maybe we chose the wrong name, that is easy to change

I'm afraid it's not easy to change, at least not if you/we care about breaking exiting code. We don't know who is accessing _symeig (e.g. Varun pointed out on the mailing list that he is using it.). It's a private field, so we can actually rename it - people must be prepared for this, right? Still, it wouldn't be so nice. I'd suggest to do the renaming in this manner:
Have `_sfa_solver` as a replacement for `_symeig`, while keeping `_symeig`. `_symeig` continues as being a switch for the raw symeig solving component (I suppose there are plenty eig and symeig routines, so this isn't so bad to have, e.g. there are GPU based solvers etc).
Then by default we init `self._sfa_solver = self._symeig` and use `_sfa_solver` in `_stop_training`.
This would introduce precise naming while not breaking existing code that uses `_symeig`.

Stewori: One more thing:
`_sfa_solver` sould have the same signature like `_symeig` I suppose, but can we slightly rename the `range` parameter? It is currently hiding the range function.

Stewori:
I created a PR with an alternative solution to this issue, based on symeig/eigh rather than svd:
https://github.com/mdp-toolkit/mdp-toolkit/pull/31

I haven't done a benchamrk yet. Telling if eigh is actually better than svd is the next ToDo.

mdp/nodes/sfa_nodes.py
Stewori: It appears to me like numx' svd actually returns U, s, V of full size, but containing some areas of zeros to represent the rank deficit. So I suppose there should be done some slicing here to s and U, see [this draft](https://github.com/mdp-toolkit/mdp-toolkit/pull/31/files#diff-e0a5be9e597a4c5d554db9950e4155edR364).

mdp/hinet/layer.py
```diff
@@ -261,6 +261,12 @@ def __init__(self, node, n_nodes=1, dtype=None):
         super(CloneLayer, self).__init__((node,) * n_nodes, dtype=dtype)
         self.node = node  # attribute for convenience
 
+    def _execute(self, x, *args, **kwargs):                                               ### VARUN's fix
+        n_samples = x.shape[0]                                                            ### VARUN's fix
+        x = x.reshape(n_samples * x.shape[1] / self.node.input_dim, self.node.input_dim)  ### VARUN's fix
+        y = self.node.execute(x)                                                          ### VARUN's fix
+        return y.reshape(n_samples, self.output_dim)                                      ### VARUN's fix
+
```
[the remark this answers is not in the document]
fabschon: No. This is a change that Varun introduced some time ago and it should be part of hist latest PR which has been delayed for some reason.

mdp/nodes/sfa_nodes.py
fabschon: We already had a case where we ran into a function that behaved differently from the numpy-default we expected. I'd switch the whole thing to straight up numpy; maybe I can find the time to do it.

mdp/nodes/sfa_nodes.py
```diff
@@ -1,12 +1,25 @@
-from builtins import str
+
+#====================================================================[ header ]
+
 from builtins import range
 __docformat__ = "restructuredtext en"
 
+import numpy as np
+import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
fabschon: Why can there not be a plot here..? It doesn't require a screen (which I would agree would be a problem) and simply provides an additional form of debug output? In addition, it is only triggered when the program is about to crash anyway; during normal operations there's no plotting.

mdp/nodes/sfa_nodes.py
fabschon: I would remove the `self._symeig` member and replace it with something like `self._eig_solve` which would be a parameter of the `__init__()` constructor, default to `eig_solve='symeig', and accept `'SVS'` and/or any other solutions on offer (e.g. Stefan's alternative).

To me this should be a nice compromise between your suggestions.

mdp/nodes/sfa_nodes.py
fabschon: One problem we'd like to address is that if `SFANode.stop_training()` crashes then there should be more information provided to help (especially) newcomers understand what the problem is without having to figure it out themselves. The moment it crashes the node should provide what is needed to address the problem.

I can see how encapsulating the diagnostics in an optional method call is cleaner, but I wouldn't see the current solution as problematic? (Though I would agree that the here proposed mechanism is probably not ideal; right now I primarily want *some* solution in the code which we can still improve along the way.)

mdp/nodes/sfa_nodes.py
fabschon: As above: I agree this is not ideal but I don't see the benefits of a diagnostic function over a simple flag?

mdp/nodes/sfa_nodes.py
fabschon: @pberkes No. In fact I've been asking around for someone more comfortable with the math to take a look at this SVD-based method to make sure this does what we want it to do. I have, however, compared the results of the SVD solution to the _symeig solution (when the latter runs successfully) and the final plots of my experiments are essentially identical.

mdp/nodes/sfa_nodes.py
fabschon: Sorry if that's a stupid question, but Python does not have constants..?

mdp/hinet/layer.py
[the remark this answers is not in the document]
otizonaizit: Varun's PR (PR #25) is waiting for @varunrajk to have some time to implement some changes…

mdp/nodes/sfa_nodes.py
otizonaizit: Well, no, the idea of `mdp.numx` was to be able to use `numpy` or `scipy` transparently. I.e. if you have only `numpy` installed, then fine, if you have `scipy` use the "better" routines available there. Note that even @Stewori noticed that the SVD implementation is way faster in `scipy` than in `numpy`. The very same arguments holds for most of the linear algebra routines in `numpy`/`scipy`. What may be obsolete is the idea of not requiring `scipy` by default. This would mean make `scipy` a hard dependency. I am fine with that, if you think it's worth it. But note that `numpy` is practically available pre-installed by default everywhere, so if you don't care about performance you can now use MDP without pulling in all of `scipy`. But, again, YMMV…

mdp/nodes/sfa_nodes.py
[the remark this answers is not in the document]
otizonaizit: But this mean pulling in matplotlib as a dependency for MDP. I don't think it's worth it. Why can't the plotting be moved somewhere outside of the node? Why plotting at all? I mean, all the necessary variables are available, so users can plot whatever they want, no?

mdp/nodes/sfa_nodes.py
otizonaizit: But why break backwards compatibility just for the sake of consistency? Just leave the name `_symeig` alone and replace its value with whatever solver you like. Regarding the `range` parameter, well, again, I don't think it's worth it to break backwards compatibility. The name comes from the parameter name in the LAPACK routine. But I don't see the problem with the `range` function. The name is overriden only in the function scope, and even there you still have  `__builtins__.range` if you really need it, no?

mdp/nodes/sfa_nodes.py
otizonaizit: Because it allows to encapsulate user code within a `try/except` without polluting the node code with diagnostics. The less flags you add ot the node, the less surprising it is to use it. Having a flag that changes the return value is considered bad practice in Python. I know that in this case the result is an Exception, but still, it seems to violate some basic code cleaness principle. If the diagnostics are happening in a different module, you have more freedom there to do whatever, without affecting existing use of the node in someone else's code.

mdp/nodes/sfa_nodes.py
otizonaizit: I think @pberkes meant to assign the string to a variable, e.g. `message="""asdasdsad"""`.

mdp/nodes/sfa_nodes.py
fabschon: That makes a lot of sense. However, we already had one case where a `scipy`method differed enough from its `numpy` equivalent to cause erroneous behavior. I'd definitely be in favor of picking one or the other.
Making `scipy` a hard dependency for a machine learning library should be fine?

mdp/nodes/sfa_nodes.py
pberkes: Yes there are no real constants, but usually people use variables written in `CAPITAL_LETTERS` at the start of the module to simulate constants. This one could be called `_SINGULAR_MATRIX_ERROR_MSG`

mdp/nodes/sfa_nodes.py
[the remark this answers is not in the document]
fabschon: Ah, the additional dependency is a good point; I agree that that's not really worth it. Plotting outside the node would still require `matplotlib` though, or do you mean "outside" as in "outside of MDP by the users themselves"?

mdp/nodes/sfa_nodes.py
fabschon: Right, backwards compatibility should definitely not be broken. In that case using `_symeig` to hold / point to whatever solver is chosen seems a good solution.

Just for the sake of argument though, how would renaming `_symeig` break backwards compatibility.. it's a private member that shouldn't be accessed from the outside either way?

mdp/nodes/sfa_nodes.py
fabschon: Right. I don't have a strong opinion here, so I'd be okay with either. However, I do feel that an additional diagnostics function is not ideal either. The basic idea behind these changes was to make the node mode accessible to other neuroscience groups, which also includes people who write less code on a daily basis.

If something breaks (which right now happens fairly often due to the eigenvalue issue) the default behavior should thus give you all the information you need to solve the problem; ideally right out of the box and without having to wrap your call int a `try` block.
I feel like there should be a cleaner way of doing this than both of the solutions proposed so far.

mdp/nodes/sfa_nodes.py
fabschon: Right, that's an easy fix.

mdp/nodes/sfa_nodes.py
otizonaizit: What we are doing right now can be understood in `mdp/utils/__init__.py`. We wrote a little compatibility layer between `numpy` and `scipy` queivalent functions, so that we are sure we use internally a consistent interface. We have `mdp.utils.svd` for example. Look also at the casting issues we address there (`numpy` casts to double even if the inputs are single in several functions). If the difference in results between a `numpy` function and a `scipy` one are more than just cosmetic, then this is bug in `numpy` or `scipy` and needs to be reported there, I think.

mdp/nodes/sfa_nodes.py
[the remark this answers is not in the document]
otizonaizit: It could be an additional module within MDP, or even an externally developed module, that's your pick :)

mdp/nodes/sfa_nodes.py
otizonaizit: Well, we have seen that Varun was using the private attribute already, no? So, people were using it even if it was a private attribute. Maybe we should have made it public from the beginning, I don't know. Nothing stops us from creating a new public property `eig_solver` or whatever, and then automatically set the `_symeig` attribute when `eig_solver` is changed, so that you get a nicer and public interface for the future, without breaking old code?

mdp/nodes/sfa_nodes.py
otizonaizit: Yes, and that's why a better error message is a good starting point. We could have a whole new chapter in the SFA documentation online, with code examples, about this issue. I think the problem is complicated enough that there is no "automatic" solution that would save the user from the hard job of understading why SFA is failing… We can simply help the user find her way by having better documentation.

mdp/nodes/sfa_nodes.py
fabschon: Right, I don't remember the details of what exactly the problem was that we had, so not sure how much of an issue this actually is. Changing the code to explicitly use either `numpy` or `scipy` would be subject to another PR anyway, so I'll just change it back to `numx` and we can think about changing this later should we feel it's necessary.

mdp/nodes/sfa_nodes.py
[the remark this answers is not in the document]
fabschon: If it was a (new) module within MDP we'd still introduce `matplotlib` as an additional dependency though, no?

mdp/nodes/sfa_nodes.py
fabschon: Hm. If we get this (or @Stewori's) PR through before Varun's, then he'll rebase it anyways and it won't matter what we call it. Of course someone else might have used the `_symeig` from the outside but.. well, they shouldn't have, as it's marked as a private attribute. Maybe someone else can pitch in on this, if there's a consensus to preserve private attributes like this then I'd be okay with that.

otizonaizit:
>If it was a (new) module within MDP we'd still introduce `matplotlib` 
>as an additional dependency though, no?
Yes, so probably an example in the docs is better than a new module?

otizonaizit:
>Hm. If we get this (or @Stewori's) PR through before Varun's, then 
>he'll rebase it anyways and it won't matter what we call it. Of course 
>someone else might have used the `_symeig` from the outside but.. well, 
>they shouldn't have, as it's marked as a private attribute. Maybe 
>someone else can pitch in on this, if there's a consensus to preserve 
>private attributes like this then I'd be okay with that.
I'd strive to keep backwards compatibility, if that does not make our 
lives impossible. You never know who is going to be hurt when you break 
it…

mdp/nodes/sfa_nodes.py
fabschon: Absolutely! In fact, a sweet intro and tutorial on SFA is already on the list for Nils here, who's working on updating the MDP documentation :)

However, in terms of making SFA more accessible to newcomers I think we can agree that the default behavior should look like this:

* User packs data into SFA, which either successfully completes training or runs into the EV issue.
* In case of the latter **(a)** a clear error message should point out the flag to switch over to a more robust solver, and **(b)** enough information should be presented immediately to explain what is going on (maybe even a link the tutorial once it's up). Regarding the latter, Laurenz has expressed he'd like an error message to include as much information as possible, so people have a chance of understanding what the problem is and what they can do about it (both in terms of their data and what options MDP offers, i.e., optional solver-alternatives).

fabschon:
> Yes, so probably an example in the docs is better

I like that idea! We can even provide the code in a `tutorials` subfolder within MDP without it being used when importing the library.

> I'd strive to keep backwards compatibility

I agree, though I keep thinking that if there's a time when such a change hurts the least, then it'd be right now. But oh well ;)

otizonaizit:
>* User packs data into SFA, which either successfully completes 
>training or runs into the EV issue.
>* In case of the latter **(a)** a clear error message should point out 
>the flag to switch over to a more robust solver, and **(b)** enough 
>information should be presented immediately to explain what is going on 
>(maybe even a link the tutorial once it's up). Regarding the latter, 
>Laurenz has expressed he'd like an error message to include as much 
>information as possible, so people have a chance of understanding what 
>the problem is and what they can do about it (both in terms of their 
>data and what options MDP offers, i.e., optional solver-alternatives).
Yep, I agree with everything, just not say "more robust solver" until we 
have demonstrated that the alternative solvers are indeed more robust at 
solving the problem, and not just delivering "some" results, which may 
turn out to be wrong. To reiterate my point once again: I may be 
mistaken, but I think that when EV fails, the results returned by SVD 
are not solving the original SFA problem.

pberkes:
Regarding testing reliability of the SVD solver, I suspect something like this should work: create an example which is a linear mixture of N slow signals and one signal with huge derivatives (N=3 or 4 might be sufficient). The derivatives should be huge enough that the `symeig` solver gives up. Run the SVD solver and show that you get back the slow signals.

Stewori:
I think we can close this PR, as most of it is covered now by [symeig_semidefinite module](https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/utils/symeig_semidefinite.py).
If there is need to still add the analysis part (maybe externally of the SFA node), this should be discussed in a fresh PR anyway.

fabschon:
Oh right, of course. For some reason I thought it got closed automatically when 
your was accepted into the system. 

-- f

PR #31, opened 2017-09-02
Sfa rank deficit solver

Stewori:
Alternative solution to https://github.com/mdp-toolkit/mdp-toolkit/pull/30, based on symeig/eigh rather than svd. This solution is mathematically equivalent to the svd variant, but

- allows control over the threshold when an eigenvalue is considered zero
- uses eigh/symeig rather than svd. If this is a benefit depends on the underlying eigh vs svd backend.

Todo: Benchmark svd vs eigh, let's say regarding scipy default implementations.


Stewori:
Now I included the SVD solving method from https://github.com/mdp-toolkit/mdp-toolkit/pull/30 for better comparability.
Additionally it occurred to me that the issue can be solved without any (significant) performance loss compared to the ordinary plain symeig based approach by applying some moderate regularization to the covariance matrix. See the [freshly added regularization based solver](https://github.com/mdp-toolkit/mdp-toolkit/pull/31/files#diff-e0a5be9e597a4c5d554db9950e4155edR254).
I augmented Fabian's profiling script to include the solvers from this PR (feel free to try):

```
"""
Script to profile the different alternatives of handling the eigenvalue issue
in the SFA training procedure. For testing purposes only (non expanded) random
noise is used; we're not interested in the results themselves, only how fast
and expensive they are to compute.
 
To use, run as follows:
   $ mprof run --python python2 profiling.py
   $ mprof plot
 
-> Time usage is printed to the shell.
-> Memory usage is collected in a .dat file plot-able via 'mprof plot'
"""

#--------------------------------------------------------------------[ header ]

import gc            # garbage collector interface (for manual triggering)
import timeit
import numpy as np

import mdp
from mdp.utils import symeig

#----------------------------------------------------------------------[ util ]

def produceGarbage( size ):
    # create a random positive-semidefinite matrix (multiply some random matrix w/ its transpose)
    print 'Generating garbage..'
    garbage = ( np.random.randn( size, size ) * 255 )
    garbage = np.dot( garbage, garbage.T )
    print 'Storing garbage in memmap..'
    memmap    = np.memmap( 'garbage_memmap', dtype=np.float32, mode='w+', shape=garbage.shape )
    memmap[:] = garbage[:]
    del memmap # flush data to disk and re-open in readonly mode
    memmap = np.memmap( 'garbage_memmap', dtype=np.float32, mode='r', shape=garbage.shape )
    print 'Garbage ready for use.'
    return memmap

#--------------------------------------------------------------------[ symeig ]

@profile
def symeig_test( cov_mtx, dcov_mtx, rng=(1,32) ):
    d, sf = symeig( dcov_mtx, cov_mtx, range=rng, overwrite=True )
    return d, sf

#-----------------------------------------------------------------------[ svd ]

@profile
def svd_test( cov_mtx, dcov_mtx, rng=(1,32) ):
    # SVD method
    U, s, _ = np.linalg.svd( cov_mtx )
    X1 = np.dot( U, np.diag(1.0 / s ** 0.5) )
    X2, _, _ = np.linalg.svd( np.dot(X1.T, np.dot(dcov_mtx, X1)) )
    E = np.dot(X1, X2)
    e = np.dot(E.T, np.dot(dcov_mtx, E)).diagonal()
    # sort eigenvalues & -vectors in ascending order and return
    e = e[::-1]
    E = E.T[::-1].T
    return e[rng[0] - 1:rng[1]], E[:, rng[0] - 1:rng[1]]

#-----------------------------------------------------------------------[ rd ]

mult = np.dot
_rank_threshold = 0.000001

@profile
def pca_test(cov_mtx, dcov_mtx, rng=(1, 32)):
    eg, ev = symeig(cov_mtx)#, None, True, turbo, None, type, overwrite)
    off = 0
    while eg[off] < _rank_threshold:
        off += 1
    eg = 1/np.sqrt(eg[off:])
    ev2 = ev[:, off:]
    ev2 *= eg
    S = ev2

    white = mult(S.T, mult(dcov_mtx, S))
    sfa_eg, sfa_ev = symeig(white)#, None, True, turbo, None, type, overwrite)
    sfa_ev = mult(S, sfa_ev)
    if rng is None:
        # self.d, self.sf
        return sfa_eg, sfa_ev
    else:
        # self.d, self.sf
        return sfa_eg[rng[0]-1:rng[1]], sfa_ev[:, rng[0]-1:rng[1]]

#-----------------------------------------------------------------------[ reg ]

@profile
def reg_test(B0, A, rng=(1, 32)):
    # apply some regularization...
    #B += 1e-14*np.eye(B.shape[0])
    # This saves some memory of np.eye
    B = B0.copy()
    Bflat = B.reshape(B.shape[0]*B.shape[1])
    idx = np.arange(0, len(Bflat), B.shape[0]+1)
    Bflat[idx] += 1e-14
    eg, ev = symeig(A, B) #, True, turbo, None, type, overwrite)
    m = np.abs(np.sqrt(np.sum(ev * np.dot(B, ev), 0))-1)
    off = 0
    while m[off] > 1e-12:
        off += 1
    if off > 0:
        #self.rank_deficit = off
        eg = eg[off:]
        ev = ev[:, off:]
    if rng is None:
        # self.d, self.sf
        return eg, ev
    else:
        # self.d, self.sf
        return eg[rng[0]-1:rng[1]], ev[:, rng[0]-1:rng[1]]

#======================================================[ run profiling checks ]

def main():

    # library info
    print 'Using numpy version,', np.__version__, '(@ %s)' % (  np.__path__[0] )
    print 'Using mdp version',   mdp.__version__, '(@ %s)' % ( mdp.__path__[0] )
 
    # parameters
    runs_per_method = 10
    garbage_size    = 2000
 
    # setup
    garbage = produceGarbage( garbage_size )
    methods = [ symeig_test,
                reg_test,
                svd_test,
                pca_test   ]
    time = np.zeros([len(methods), runs_per_method])
 
    # profiling runs
    for i_m, m in enumerate(methods):
        print '\nRunning', m.__name__
        for i in range(runs_per_method):
            print '    Run %d/%d..'% ( i+1, runs_per_method ),
            ping = timeit.default_timer()
            _, _ = m( garbage, garbage )
            time[i_m,i] = timeit.default_timer() - ping
            print 'done. (' + str(time[i_m,i]) + ')'
        print '    Avg. time (excluding first run):', np.mean( time[i_m,1:] )
        print '    Min. time (excluding first run):', np.min ( time[i_m,1:] )
        gc.collect()
 
    print 'All done.'
 
main()
```

Stewori:
The failing CI check seems to be unrelated to my changes. Maybe it's just some bad coincidence in a random-data-based test...?

Stewori:
After running the tests of my machine several times, I can confirm that `test_CCIPCANode` sometimes fails, which is also what the CI complains about. Maybe some threshold should be adjusted for that test...?

Stewori:
As a side effect it is now possible to rerun a failed stop_training, e.g. after setting a different `_sfa_solver`:
https://github.com/mdp-toolkit/mdp-toolkit/pull/31/files#diff-e0a5be9e597a4c5d554db9950e4155edR172
https://github.com/mdp-toolkit/mdp-toolkit/pull/31/files#diff-e0a5be9e597a4c5d554db9950e4155edR179

This is used/tested in https://github.com/mdp-toolkit/mdp-toolkit/pull/31/files#diff-014aa0eba7c7d99e0021224acf743c50R124

Stewori:
Here are the results of profiling.py on my laptop (rd is pca-based method, reg is regularization-based method, symeig is ordinary SFA):

```
mprof run --python python2 profiling.py
mprof: Sampling memory every 0.1s
running as a Python program...
Using numpy version, 1.11.0 (@ /usr/lib/python2.7/dist-packages/numpy)
Using mdp version 3.5 (@ mdp)
Generating garbage..
Storing garbage in memmap..
Garbage ready for use.

Running symeig_test
    Run 1/10.. done. (3.88429999352)
    Run 2/10.. done. (3.82046508789)
    Run 3/10.. done. (3.82220387459)
    Run 4/10.. done. (3.8009250164)
    Run 5/10.. done. (3.8292350769)
    Run 6/10.. done. (3.83064103127)
    Run 7/10.. done. (3.87589287758)
    Run 8/10.. done. (3.86247801781)
    Run 9/10.. done. (3.86444497108)
    Run 10/10.. done. (3.82919883728)
    Avg. time (excluding first run): 3.83727608787
    Min. time (excluding first run): 3.8009250164

Running reg_test
    Run 1/10.. done. (4.0262389183)
    Run 2/10.. done. (4.02308511734)
    Run 3/10.. done. (4.01411294937)
    Run 4/10.. done. (4.01668715477)
    Run 5/10.. done. (4.01696610451)
    Run 6/10.. done. (4.1180369854)
    Run 7/10.. done. (4.03211092949)
    Run 8/10.. done. (4.04859995842)
    Run 9/10.. done. (4.03108692169)
    Run 10/10.. done. (4.02839398384)
    Avg. time (excluding first run): 4.03656445609
    Min. time (excluding first run): 4.01411294937

Running svd_test
    Run 1/10.. done. (24.9563009739)
    Run 2/10.. done. (26.8296639919)
    Run 3/10.. done. (27.1757938862)
    Run 4/10.. done. (25.7094230652)
    Run 5/10.. done. (25.7800080776)
    Run 6/10.. done. (25.7114269733)
    Run 7/10.. done. (27.26694417)
    Run 8/10.. done. (25.9730739594)
    Run 9/10.. done. (25.6829259396)
    Run 10/10.. done. (25.8364369869)
    Avg. time (excluding first run): 26.2184107833
    Min. time (excluding first run): 25.6829259396

Running rd_test
    Run 1/10.. done. (9.7182328701)
    Run 2/10.. done. (9.56088805199)
    Run 3/10.. done. (9.5909819603)
    Run 4/10.. done. (10.4158859253)
    Run 5/10.. done. (9.78671598434)
    Run 6/10.. done. (9.66823410988)
    Run 7/10.. done. (9.75410580635)
    Run 8/10.. done. (9.69572997093)
    Run 9/10.. done. (9.70559000969)
    Run 10/10.. done. (9.72135686874)
    Avg. time (excluding first run): 9.76660985417
    Min. time (excluding first run): 9.56088805199
All done.
```

Given the huge gap to [Fabian's timing of svd](https://github.com/mdp-toolkit/mdp-toolkit/pull/30) it occurred to me that my NumPy might be using an odd svd implementation. If I adjust `profiling.py` to use svd from SciPy (symeig is taken from mdp, i.e. also from SciPy I suppose) I get:

```
mprof run --python python2 profiling.py
mprof: Sampling memory every 0.1s
running as a Python program...
Using numpy version, 1.11.0 (@ /usr/lib/python2.7/dist-packages/numpy)
Using mdp version 3.5 (@ mdp)
Generating garbage..
Storing garbage in memmap..
Garbage ready for use.

Running symeig_test
    Run 1/10.. done. (4.2249109745)
    Run 2/10.. done. (4.06828308105)
    Run 3/10.. done. (3.86261892319)
    Run 4/10.. done. (3.84835410118)
    Run 5/10.. done. (3.86517906189)
    Run 6/10.. done. (3.87726306915)
    Run 7/10.. done. (3.87809705734)
    Run 8/10.. done. (3.87279510498)
    Run 9/10.. done. (3.88489317894)
    Run 10/10.. done. (3.87167406082)
    Avg. time (excluding first run): 3.89212862651
    Min. time (excluding first run): 3.84835410118

Running reg_test
    Run 1/10.. done. (4.05949997902)
    Run 2/10.. done. (4.05225896835)
    Run 3/10.. done. (4.04913282394)
    Run 4/10.. done. (4.05109000206)
    Run 5/10.. done. (4.03543591499)
    Run 6/10.. done. (4.0659558773)
    Run 7/10.. done. (4.03354811668)
    Run 8/10.. done. (4.04254817963)
    Run 9/10.. done. (4.02322101593)
    Run 10/10.. done. (4.10549283028)
    Avg. time (excluding first run): 4.0509648588
    Min. time (excluding first run): 4.02322101593

Running svd_test
    Run 1/10.. done. (13.917580843)
    Run 2/10.. done. (13.7519381046)
    Run 3/10.. done. (13.6480340958)
    Run 4/10.. done. (13.7324337959)
    Run 5/10.. done. (13.6220281124)
    Run 6/10.. done. (13.727959156)
    Run 7/10.. done. (13.6534600258)
    Run 8/10.. done. (13.616038084)
    Run 9/10.. done. (13.7160699368)
    Run 10/10.. done. (13.8770840168)
    Avg. time (excluding first run): 13.7050050365
    Min. time (excluding first run): 13.616038084

Running rd_test
    Run 1/10.. done. (9.49105000496)
    Run 2/10.. done. (9.47291398048)
    Run 3/10.. done. (9.53901815414)
    Run 4/10.. done. (9.51795816422)
    Run 5/10.. done. (9.62437295914)
    Run 6/10.. done. (9.56001996994)
    Run 7/10.. done. (9.54202985764)
    Run 8/10.. done. (9.47071099281)
    Run 9/10.. done. (9.59698605537)
    Run 10/10.. done. (9.52712488174)
    Avg. time (excluding first run): 9.53901500172
    Min. time (excluding first run): 9.47071099281
All done.
```

This is much better, but still the regularization-based method is the fasted. I'd even recommend to make it the default for SFA overall, given its better robustness, but that's certainly not my decision....

mdp/nodes/sfa_nodes.py
```diff
@@ -214,6 +253,137 @@ def get_eta_values(self, t=1):
...
+        (even if overwrite=False), but the changes are negligible.
+        """
+        if type != 1:
+            raise ValueError('Only type=1 is supported.')
+        # apply some regularization...
+        # The following is equivalent to B += 1e-14*np.eye(B.shape[0]),
+        # but works more in place, i.e. saves memory consumption of np.eye().
+        Bflat = B.reshape(B.shape[0]*B.shape[1])
+        idx = numx.arange(0, len(Bflat), B.shape[0]+1)
+        Bflat[idx] += 1e-12
```
otizonaizit: I don't think this is a good idea. You are unconditionally adding a value to the whole matrix. The eigenvalues and the eigenvectors are not invariant under translation, so you are effectively changing the solution. Because the value is small in absolute value, you would normally not notice the problem. But SFA is interested in the **smallest** eigenvalues, so you are interfering with exactly those eigenvalues we care most. There is also a problem with numerical stability. If SFA is applied in a polynomial space (say with degree higher than 3), then the eigenvalues will have a dynamic range which is often quite big, with a ratio betwen the biggest and the smallest eigenvalue which sometimes can exceed double numerical precision (1e-13). In my experience these are the cases where the standard eigensolver has problems, because of accumulation of numerical errors. By doing this regularization you are putting the problem under the carpet somehow, but the numers you get have nothing to do with the original problem… I have the feeling that we need a test case which is not fictional, and we should then compare the different approaches.

otizonaizit: In general I have to say that before going on with this approach we should check that the SVD and the regularization approach are mathematically **and** computationally equivalent to the generalized eigenvalues approach. I am still not completely sure about it, especially in the numerically instable cases, where we care exactly about those eigenvalues that SVD and the regularization are going to discard… That is why until now the preferred and suggested solution to the problem has been to use PCA before SFA: this allows you to control exactly what is begin neglected, and it does that in a space where "throwing away" dimensions is safe.

fabschon:
I've been running the benchmark script with your alternatives as well:
```
Running symeig_test
    Avg. time (excluding first run): 11.5670145088
    Min. time (excluding first run): 11.2888891697

Running reg_test
    Avg. time (excluding first run): 20.5124979019
    Min. time (excluding first run): 20.309472084

Running svd_test
    Avg. time (excluding first run): 11.28693072
    Min. time (excluding first run): 11.0398859978

Running pca_test
    Avg. time (excluding first run): 22.9437248177
    Min. time (excluding first run): 22.7159218788
All done.
```

Which shows:
`pca` **>** `reg` **>** `symeig` **>** `svd`.

In contrast, your (@Stewori)  timings show:
`svd` **>** `rd` **>** `reg` **>** `symeig`.

Which is, well, interesting. What kind of math backend are you using? (You can check with np.__config__.show(); here's [mine](https://pastebin.com/tjUVwU2w).) And CPU? (I'm running on an i7-2600K (3.40GHz).)

Can you add the `rd` solution to the script as well? I'll also ask around here for more people to run the script so we get a better idea of how to interpret that discrepancy in the benchmark results.

otizonaizit:
For the benchmarking I would suggest **not** to use a random matrix, but construct a matrix that exhibits the problems and work with that. The speed my well depend on how "hard" it is to get the eigenvalues. But why are we benchmarking anyway? If everyone agrees that the symeig solution is the mathematically and numerically correct one in almost all cases, than for those cases where it breaks speed is probably not the main issue. I guess if you are in need of SVD by default, than there is something wrong with the data?

fabschon:
We wanted to check if there are significant differences in time / memory consumption when using the different methods of dealing with the eigenvalue problem. Ideally we'll choose the fastest method as the default behavior and the fastest rank deficit solution in case of problematic data.

otizonaizit:
But I keep on thinking that changing the default can only be done once 
it has been demonstrated that the new method is mathematically and 
numerically equivalent to the generalized eigenvalue problem one…

fabschon:
Right. As I said, I've run through a couple of my experiments which use a hierarchical SFA network, and computed the same results with both `svd()` and `symeig()`. I'd also like a more mathematically inclined person to look at the SVD solution (I can do it too, it'll just take longer). Other than that, what would you like to see to make sure a possible replacement for the default behavior works as intended?

otizonaizit:
>Right. As I said, I've run through a couple of my experiments which use 
>a hierarchical SFA network, and computed the same results with both 
>`svd()` and `symeig()`. I'd also like a more mathematically inclined 
>person to look at the SVD solution (I can do it too, it'll just take 
>longer). Other than that, what would you like to see to make sure 
>a possible replacement for the default behavior works as intended?
I'd like to have Laurenz to vouch for the SVD solution, after having 
consider the numerical issues (again, it is not just the  mathematics, 
it is also a matter of the behaviour of SVD when the standard EV fails: 
is SVD doing the right thing there?). I am sure you can compute the same 
results with SVD or symeig when the data is OK, that's automatic, 
because SVD is equivalent to EV when the matrix has full rank and the 
eigenvalues are far from zero. The question is what happens when this 
assumption is not met. Are the two methods still equivalent? I am not 
sure about this.

mdp/nodes/sfa_nodes.py
Stewori: > You are unconditionally adding a value to the whole matrix.

Only to the diagonal, so this procedure enforces full rank.

> The eigenvalues and the eigenvectors are not invariant under translation, so you are effectively changing the solution. Because the value is small in absolute value, you would normally not notice the problem.

I think the regularization of order 1e-12 should not perturb the eigenvalues significantly. However you will get a slightly (i.e. measurably) different eigenvector/extraction matrix. This is actually okay and expectable, because the optimal extraction matrix is not unique if the input data contains linear redundancy. [testSFANode_rank_deficit()](https://github.com/mdp-toolkit/mdp-toolkit/pull/31/files#diff-014aa0eba7c7d99e0021224acf743c50R80) already ensures that the resulting extraction matrix is of equivalent quality.

> But SFA is interested in the smallest eigenvalues, so you are interfering with exactly those eigenvalues we care most.

The implementation distinguishes small eigenvalues introduced by regularization from actual slow features based on the covariance matrix rather than the dcov matrix (see [this line and following](https://github.com/mdp-toolkit/mdp-toolkit/pull/31/files#diff-e0a5be9e597a4c5d554db9950e4155edR285)).
With a covariance matrix B and extraction matrix V the identity `V.T B V == I` [should hold](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.linalg.eigh.html#scipy.linalg.eigh). Actual rank deficit of B (only fixed by regularization) leads to a significant violation of this equation (Dimensions belonging to rank deficit differ from 1 by a threshold ~1e-4, which is really significant).
I believe this is a robust way to filter away purely regularization-induced almost-zero eigenvalues while not significantly changing actual slow-feature-induced ones.

> I have the feeling that we need a test case which is not fictional, and we should then compare the different approaches.

I absolutely agree. Suggestions for a proper test case? Would it be sufficient to use [the data that ordinary SFA is tested on](https://github.com/Stewori/mdp-toolkit/blob/e887365a00efe14307099e159b82231407fc7028/mdp/test/test_SFANode.py#L10) (with some redundancy added)?

Stewori:
Agreed that we should test on non-random data.

> Can you add the rd solution to the script as well?

That's pca_test. I renamed it, because the shortcut `rd` stems from the time where it was the only rank deficit solver in my PR.

Given the strongly altering timings between the backends I'd suggest to add all approaches so it is easy to switch/check out the best one for each particular case. Only question is how to select the default (for rank deficit solving) -- I suppose we need more testing/benchmarking for that.

I'm perfectly fine with keeping the overall SFA default approach like it is. Using a rank deficit solver as default was just a recommendation under the assumption that the regularization-based approach would be always (almost) as fast as plain symeig.

Regarding mathematical equivalence, I am rather sure that the approaches as implemented in this PR are mathematically equivalent up to numerical implementation details of the underlying backends.
[testSFANode_rank_deficit()](https://github.com/mdp-toolkit/mdp-toolkit/pull/31/files#diff-014aa0eba7c7d99e0021224acf743c50R80) shows that they are also computationally equivalent.

The only measurable difference is the extraction matrix found by the regularization-based approach, but that's expectable because as soon as the input data has linear redundancy the optimal extraction matrix is not unique any more. As the comparison of output data and eigenvalues show, the extraction matrices are equivalent in quality.

otizonaizit:
OK, Let's try to disentangle the two situations. With your tests in #31 you are testing the case where there is some linear redundancy in the data. And I agree with you that in this case the regularization, and the SVD approach for that matter, are mathematically and numerically equivalent. @pberkes suggested a test for the case where the dynamic range of the eigenvalues is big enough to create problems in `symeig`. Are the methods still equivalent in this case? I am not sure. I'd like to have a test that indeed reproducibly induces the exception in `symeig`. Starting from there we can then see what the other solvers are doing in this case and evaluate the results (indepedent of the benchmarking, which is interesting but not really important for the problem at hand).

Stewori:
@fabschon 

```
>>> import numpy as np
>>> np.__config__.show()
lapack_info:
    libraries = ['lapack', 'lapack']
    library_dirs = ['/usr/lib']
    language = f77
lapack_opt_info:
    libraries = ['lapack', 'lapack', 'blas', 'blas']
    library_dirs = ['/usr/lib']
    language = c
    define_macros = [('NO_ATLAS_INFO', 1), ('HAVE_CBLAS', None)]
openblas_lapack_info:
  NOT AVAILABLE
blas_info:
    libraries = ['blas', 'blas']
    library_dirs = ['/usr/lib']
    define_macros = [('HAVE_CBLAS', None)]
    language = c
atlas_3_10_blas_threads_info:
  NOT AVAILABLE
atlas_threads_info:
  NOT AVAILABLE
atlas_3_10_threads_info:
  NOT AVAILABLE
atlas_blas_info:
  NOT AVAILABLE
atlas_3_10_blas_info:
  NOT AVAILABLE
atlas_blas_threads_info:
  NOT AVAILABLE
openblas_info:
  NOT AVAILABLE
blas_mkl_info:
  NOT AVAILABLE
blas_opt_info:
    libraries = ['blas', 'blas']
    library_dirs = ['/usr/lib']
    language = c
    define_macros = [('NO_ATLAS_INFO', 1), ('HAVE_CBLAS', None)]
atlas_info:
  NOT AVAILABLE
atlas_3_10_info:
  NOT AVAILABLE
lapack_mkl_info:
  NOT AVAILABLE
mkl_info:
  NOT AVAILABLE
```

```
>>> import scipy
>>> scipy.__config__.show()
lapack_info:
    libraries = ['lapack']
    library_dirs = ['/usr/lib']
    language = f77
lapack_opt_info:
    libraries = ['lapack', 'blas']
    library_dirs = ['/usr/lib']
    language = c
    define_macros = [('NO_ATLAS_INFO', 1), ('HAVE_CBLAS', None)]
openblas_lapack_info:
  NOT AVAILABLE
blas_info:
    libraries = ['blas']
    library_dirs = ['/usr/lib']
    define_macros = [('HAVE_CBLAS', None)]
    language = c
atlas_3_10_blas_threads_info:
  NOT AVAILABLE
atlas_threads_info:
  NOT AVAILABLE
atlas_3_10_threads_info:
  NOT AVAILABLE
atlas_blas_info:
  NOT AVAILABLE
atlas_3_10_blas_info:
  NOT AVAILABLE
atlas_blas_threads_info:
  NOT AVAILABLE
openblas_info:
  NOT AVAILABLE
blas_mkl_info:
  NOT AVAILABLE
blas_opt_info:
    libraries = ['blas']
    library_dirs = ['/usr/lib']
    language = c
    define_macros = [('NO_ATLAS_INFO', 1), ('HAVE_CBLAS', None)]
atlas_info:
  NOT AVAILABLE
atlas_3_10_info:
  NOT AVAILABLE
lapack_mkl_info:
  NOT AVAILABLE
mkl_info:
  NOT AVAILABLE
>>> 
```

(Yes, I really should improve my configuration.)
Regarding your benchmark, the huge difference between `symeigtest` and `reg_test` puzzles me. It really shouldn't be so large, because they use the same backend function and the additional steps of reg_test are not very expensive. Maybe this is due to different values for `range` and `overwrite`.

mdp/nodes/sfa_nodes.py
```diff
@@ -68,10 +72,37 @@ class SFANode(Node):
...
+          is equivalent to
+
+             sfa = SFANode()
+             sfa._sfa_solver = sfa._rank_deficit_solver_reg
+          
     """
 
     def __init__(self, input_dim=None, output_dim=None, dtype=None,
-                 include_last_sample=True):
+                 include_last_sample=True, handle_rank_deficit=False):
```
fabschon: As mentioned before, I'd prefer to name the solutions directly, i.e., an `eig_solve` (for example) parameter than takes the values of `'symeig'`, `'svd'`, etc. and allows to directly choose the preferred method. The default can either be `'symeig'` or whatever comes out ahead during benchmark testing.

mdp/nodes/sfa_nodes.py
```diff
@@ -68,10 +72,37 @@ class SFANode(Node):
...
+          select by setting the _sfa_solver field, e.g.
+
+             sfa = SFANode()
+             sfa._sfa_solver = sfa._rank_deficit_solver_pca
+
+          Currently there are three solvers available:
+
+          _rank_deficit_solver_reg    - works by regularization
+          _rank_deficit_solver_pca - works by PCA
+          _rank_deficit_solver_svd    - works by svd
```
fabschon: We should add at least some information about how one would choose a method in case `'symeig'` fails, i.e., why choose regularization over PCA etc. We might have to wait until we have more benchmark results for this though.

mdp/nodes/sfa_nodes.py
```diff
@@ -87,6 +118,10 @@ def __init__(self, input_dim=None, output_dim=None, dtype=None,
 
         # set routine for eigenproblem
         self._symeig = symeig
+        self._sfa_solver = (self._rank_deficit_solver_reg if handle_rank_deficit
+                else self._symeig)
+        self._rank_threshold = 1e-12
+        self.rank_deficit = 0
```
fabschon: Ideally, instead of four members, we would only have a single member variable here, e.g., `_eig_solve` and move any parameters needed for the individual solvers to their respective methods. In my mind this would make the `__init__` here a lot cleaner, which would help make the SFANode easier to understand.

mdp/nodes/sfa_nodes.py
```diff
@@ -134,22 +169,25 @@ def _train(self, x, include_last_sample=None):
 
     def _stop_training(self, debug=False):
         ##### request the covariance matrices and clean up
-        self.cov_mtx, self.avg, self.tlen = self._cov_mtx.fix()
-        del self._cov_mtx
+        if hasattr(self, '_dcov_mtx'):
+            self.cov_mtx, self.avg, self.tlen = self._cov_mtx.fix()
+            del self._cov_mtx
```
fabschon: What necessitates the `hasattr()` call here..?

mdp/nodes/sfa_nodes.py
Stewori: I suppose handle_rank_deficit should be turned off by default and should be trivial for a user to be turned on (using a default rank deficit solver), i.e. without looking into docs or code for available solver names.

I thought about having e.g. the reg-solver as default name, i.e. `rank_deficit_solver=__rank_deficit_solver_reg`, but then how would we have rank deficit solving turned off by default?

mdp/nodes/sfa_nodes.py
Stewori: Then it would be impossible for an advanced user to configure them.
Also note that rank_deficit is meant as an output parameter, so an advanced user can transparently see how much rank deficit was detected.

mdp/nodes/sfa_nodes.py
Stewori: With that check it is possible to rerun a failed stop_training call with a different `_sfa_solver`, e.g. one of those supporting `rank_deficit`. Without the check, the whole training phase would need to be repeated.

mdp/nodes/sfa_nodes.py
```diff
@@ -134,22 +169,25 @@ def _train(self, x, include_last_sample=None):
...
         rng = self._set_range()
 
         #### solve the generalized eigenvalue problem
         # the eigenvalues are already ordered in ascending order
         try:
-            self.d, self.sf = self._symeig(self.dcov_mtx, self.cov_mtx,
-                                     range=rng, overwrite=(not debug))
+            self.d, self.sf = self._sfa_solver(
+                    self.dcov_mtx, self.cov_mtx, True, "on", rng,
+                    overwrite=(not debug))
```
fabschon: In my current revision this part looks as follows:
```
        # solve the generalized eigenvalue problem
        if self._eig_solve is 'symeig':
            self.d, self.sf = self._eig_solve_symeig( rng, debug )
        elif self._eig_solve is 'svd':
            self.d, self.sf = self._eig_solve_svd( rng )
```
I find this a bit cleaner than "hiding" the calls behind the non-descript `_sfa_solver()` call. To me it's okay if the different methods have different signatures, as they are, well, different methods. Forcing the same signature on all of them feels counter productive to me. While I can see how this approach enable the easy switching of methods, I'd still prefer a more explicit style that's easier to follow for newcomers facing the code.

mdp/nodes/sfa_nodes.py
```diff
@@ -159,7 +197,8 @@ def _stop_training(self, debug=False):
                            " or PCANode(svd=True)"% str(d))
                 raise NodeException(err_msg)
         except SymeigException as exception:
-            errstr = str(exception)+"\n Covariance matrices may be singular."
+            errstr = (str(exception)+"\n Covariance matrices may be singular."
+                    +"\n Try to create SFA node with handle_rank_deficit=True.")
```
fabschon: This error message should be a lot more verbose. Laurenz thinks that this error should say as much about what happens here as necessary for users to understand the problem. 

Here's the version from my PR, which might also not be ideal yet:
```
 except SymeigException as exception:
    errstr  = str(exception) + "\n Covariance matrices may be singular.\n"
    errstr += "This usually happens if there are redundancies in the (expanded) training data.\n" + \
              "There are several ways to deal with this issue:\n\n" + \
              "   - Use more data.\n\n" + \
              "   - Add noise to the data. This can be done by slotting an additional NoiseNode\n" + \
              "     in front of a troublesome SFANode. Noise levels do not have to be high.\n\n" + \
              "   - Run training data through PCA. This can be done by slotting an additional\n" + \
              "     PCA node in front of the troublesome SFANode. Use the PCA node to discard\n" + \
              "     dimensions within your data with lower variance.\n\n" + \
              "   - [ RECOMMENDED ] Use SVD by setting the SFANode flag use_svd to True. This\n" + \
              "     solves the internal SFA eigenvalue problem by using two instances of singular\n" + \
              "     value decomposition (SVD) instead of the default behavior of using the\n" + \
              "     systems's symeig() routine to solve a single, generalized eigenvalue problem.\n" + \
              "     Using SVD is slightly more expensive but more robust.\n"
    raise NodeException(errstr)
```
> Using SVD is slightly more expensive but more robust

At least this will probably have to be re-written once we have the full benchmarks of all the methods.

mdp/nodes/sfa_nodes.py
Stewori: As I perceived it, it is kind of an MDP requirement that signatures are fully compatible. I changed it to call-by-order rather than call-by-name, because some of the argument names hide global names, i.e. `type` and `range`. A former version of one of my solvers used `range` for iteration, so I wanted to be able to rename the range parameter, which wouldn't be possible in call-by-name case. Now it's not necessary any more, still it is bad style to hide global names and the call should allow implementations to choose different names. That's why call-by-order might be better here.

mdp/nodes/sfa_nodes.py
Stewori: I'm fine with having a more verbose error message, as long as it contains an easy instruction how to trivially  fix the issue.

mdp/nodes/sfa_nodes.py
fabschon: Well if the user runs into the problem, then the error message will tell her how to deal with it either way: Switch `eig_solve` to a different value (and listing the available ones) or switching `handle_rank_deficit` to `True` - however, in case of the latter, which method will be used to handle the rank deficit and how can we choose from the multiple solutions on offer..? I still feel like the method should be explicitly named (and chosen) by setting this parameter.

mdp/nodes/sfa_nodes.py
fabschon: The parameters would still be here though, just part of the respective solvers. Putting them there should not keep an advanced user from setting them, but prevent a non-advanced user to accidentally mess them up ;)

mdp/nodes/sfa_nodes.py
fabschon: Ah, I see. Hm. Is that what we want? This would modify the training procedure on the fly, without giving the user control over this. My preference would be to let it crash, and allow the user to adapt the whole procedure until a configuration is found that runs stable for their respective use case. Not sure I have a strong opinion about this though, it only seems like it might take away an option that I'd like to have control over myself -- but then again, that behavior can of course be optional.. which, however, would also make the whole node more complicated.

mdp/nodes/sfa_nodes.py
fabschon: Signatures for the same procedures, probably yes. My point would be that the different methods on offer are not the same procedures and thus shouldn't be treated as such.

mdp/nodes/sfa_nodes.py
Stewori: after a failed stop_training the node is usually garbage, isn't it? The user can still decide whether to redo the training phase on a fresh node or whether to recycle the garbage by rerunning stop_training.

fabschon: It might be useful to create backups after each successful training phase and store them in some temp folder. Say two training phases are successful, leaving behind two backup files, and the third phase crashes. The error message would tell the user where to find the backups and training can resume with different parameters for the problematic layer. That would be the same functionality (if I understand it correctly) but leave the control over what happens / how training is executed to the user.

otizonaizit: There is a system already in place to do this kind of things: look for `CheckpointFlow` and `Flow.set_crash_recovery` (all of this documented in the Flow class docstring). This is done at the flow level, because there is where you typically do training and testing nodes. Additionally, I don't understand what you mean that after a failed stop_training the node is garbage. If you used `stop_training(debug=True)` all internal attributes are stored, so you can actually re-run `stop_training` after having changed something else in the node, for example `node._symeig`(or `_eig_solver` or whatever you want to call it)

Stewori: You would need to manually restore the `_cov_mtx` and `_dcov_mtx` attributes from `cov_mtx` and `dcov_mtx`, even if running with `debug=True`, because `del self._cov_mtx` and `del self._dcov_mtx` are unconditionally executed before we know whether `stop_training()` succeeds. On another call it would crash on trying to execute `self.cov_mtx, self.avg, self.tlen = self._cov_mtx.fix()`.
While possible, this kind of manual fix requires knowledge about the node's internals and a user couldn't figure it out without looking into the code. (Things like this are the reason why I have to look into MDP code frequently, which isn't a good user experience.)

I find the backup idea and `CheckpointFlow` stuff much more advanced and complicated than adding these simple existence checks for `_cov_mtx` and `_dcov_mtx`. Really don't see why they would hurt so much or why they would reduce user control.

Stewori: Sorry, there is a typo in https://github.com/Stewori/mdp-toolkit/blob/e887365a00efe14307099e159b82231407fc7028/mdp/nodes/sfa_nodes.py#L172
Of course it should be
`if hasattr(self, '_cov_mtx'):` rather than
`if hasattr(self, '_dcov_mtx'):`.

mdp/nodes/sfa_nodes.py
Stewori: It would also take a way the option to see available parameters via inspection or the `dir` command. Additionally it would pollute the code of the solvers with existence checks and explicit default value assignments. Initializing all available parameters at one central place is IMO the cleanest and most transparent solution. IIRC adding and removing parameters after `__init__` is - while possible - considered bad style in Python OOP programming, but I might remember it wrong.

Stewori: Also note that `self._rank_threshold` is acommon parameter used by all rank deficit solvers, including [the SVD one](https://github.com/Stewori/mdp-toolkit/blob/e887365a00efe14307099e159b82231407fc7028/mdp/nodes/sfa_nodes.py#L366). So moving it to the solver code would lead to redundant code, which should also be avoided.

mdp/nodes/sfa_nodes.py
Stewori: I'm perfectly fine with extending this doc.

mdp/nodes/sfa_nodes.py
Stewori: There is another issue to consider. The solvers need access to `self` in order to retrieve `_rank_threshold` and to write `rank_deficit`. So the solver methods are bound to the SFANode instance which would complicate to hand them as function pointers.
If we would call it like
`sfa = SFANode(output_dim=50, sfa_solver=SFANode._rank_deficit_solver_reg)`.
we would have to add code for manually binding the function to self (i.e. turning it into a method). I think this is possible in Python, but I would have to find out how it is done. I suspect this kind of thing might also require a Python2 vs Python3 distinction.
What about using keywords for solver names? I remember this pattern from several libraries, including SciPy and Tensorflow.
I.e. the doc would provide a list of solver names e.g. "reg", "svd", "pca", that are acceptable values. This way we could include values "none" and "default" or "auto". "auto" would insert the solver that performs best in our benchmarks (after we find a proper config to base benchmarks reliably...), while "none" would default to plain symeig.

```
    def __init__(self, input_dim=None, output_dim=None, dtype=None,
            include_last_sample=True, rank_deficit_handler="none"):
```
`sfa1 = SFANode(output_dim=50, rank_deficit_handler="svd")`
`sfa2 = SFANode(output_dim=50, rank_deficit_handler="auto")`



Using `sfa._sfa_solver = whatever` it would still be possible to insert a fully custom solver.

mdp/nodes/sfa_nodes.py
Stewori: In https://github.com/mdp-toolkit/mdp-toolkit/pull/30#discussion_r136691665 pberkes stated that
> Nothing else in _stop_training needs to change.

I took that as a hint to avoid a case distinction inside of `stop_training`.

mdp/nodes/sfa_nodes.py
fabschon: Right, I can agree with those points. I would, however, make it more clear in the `__init__` how/where the parameters are used. I'd suggest once the code is feature complete then we take one of the upcoming Tuesdays, go through it together, and try to finalize it in person (still post it here before we hit the switch though, obviously).

mdp/nodes/sfa_nodes.py
fabschon: > Things like this are the reason why I have to look into MDP code frequently, which isn't a good user experience.

Same here (and in our office in general). I feel like this should be a fundamental guideline for any upcoming changes (in both the code and the documentation), to help alleviate this as much as we can.

As for a system to redo training phases on the fly: I'm in two minds about the whole idea; too much of such "here, let me help you"-features can quickly take away important decisions of the design process. But as long as it stays optional it shouldn't hurt either, and the `CheckpointFlow` seems to completely outsource it, too.

mdp/nodes/sfa_nodes.py
fabschon: > What about using keywords for solver names?

This is precisely what I was suggesting in my comments so that's obviously an excellent idea ;)

Except that it shouldn't be called `sfa_sover`, but something like `eig_solve` since it's solving the eigenvalue problem, not SFA itself.

mdp/nodes/sfa_nodes.py
fabschon: As I mentioned somewhere else (the number of mini-threads is becoming unwieldy) I'd suggest once the code is feature complete, we make a documentation and usability iteration of your changes in person and then post it here for (hopefully final) review.

mdp/nodes/sfa_nodes.py
fabschon: I think the point @pberkes was making is that it is easily possible to put every method behind the same signature and set the chosen approach in the `__init__`, leaving `_stop_training` completely untouched.

While this is a very efficient way of implementing it, my personal preference would be an explicit selection from the available eigenvalue solvers. (Though when we're at the point of arguing personal preferences, then we might as well go with any working solution; that's usually not time well spent.)

Stewori:
After some refactoring and adding more testing (also with more nasty data) and LDL-based approach, from my perspective this PR is now ready for merging. The doc is currently not specifically formatted, but I'd prefer to postpone this until a clear decision and solution for doc style is found.

I'd happily apply the merge by myself, but wanted to give another opportunity for eventual review or comments.

Stewori:
After a final in-person meeting with @fabschon and Laurenz and given that there were no further comments on this PR for some weeks I will merge it now. Tests were designed carefully and test as strict as numerical stability of the underlying methods allow. Given the inconsistent results on benchmarking, all methods are included, so the user can select what works best for her.
A collection of rank deficit robust symeig solvers is added as mdp/utils/symeig_semidefinite.py. These form the backbone of the new SFA functionality and can later be adopted for use by other nodes as well.
I hope this suits everyone's needs and improves usability of SFA!

Issue #32, opened 2018-01-15
Cannot import latest mdp using python 3


AlbertoEsc:
The latest version of mdp is imported properly when using python 2, but when I try to import mdp using python 3 I get the following exception:

> $ python3 -c "import mdp"
> Traceback (most recent call last):
>  File "<string>", line 1, in <module>
>  File "/home/escalafl/MDP_Fork_AlbertoEsc/mdp-toolkit/mdp/__init__.py", line 131, in <module>
>    from . import utils
>  File "/home/escalafl/MDP_Fork_AlbertoEsc/mdp-toolkit/mdp/utils/__init__.py", line 36, in <module>
>    from .symeig_semidefinite import (symeig_semidefinite_reg,
>  File "/home/escalafl/MDP_Fork_AlbertoEsc/mdp-toolkit/mdp/utils/symeig_semidefinite.py", line 58, in <module>
>    from _symeig import SymeigException
>ImportError: No module named '_symeig'

Apparently this is related to (the absence of) relative imports in python3 and the easiest fix is to change "from _symeig import SymeigException" to "from ._symeig import SymeigException" (an extra period) in symeig_semidefinite.py

Stewori:
Oh, sorry, I guess this was my fault. Forgot to test symeig_semidefinite with Python3. I will fix it as soon as possible.

Stewori:
Fixed in https://github.com/mdp-toolkit/mdp-toolkit/pull/33. While testing on Python 3 I encountered 58 failing tests. And I thought mdp was already Python 3 proof... :(
Or do I have some misconfiguration?

Stewori:
Maybe it's because of an old NumPy version in my Python 3 setup...

PR #33, opened 2018-01-17
Fix symeig_semidefinite for Python 3

Stewori:
Fix symeig_semidefinite for Python 3
and renamed ``_symeig._assert_eigenvalues_real_and_positive`` to ``_symeig._assert_eigenvalues_real``, because it actually does not check for positivity, so the original name was misleading. This is internal API of _symeig.py and thus should not break anything (and tests confirm it doesn't). Ther was another issue with Python3 and ``_symeig`` that could have lead to nan values in reults instead of raising an exception. This might be due to older NumPy, older SciPy or Python3 (cannot really tell), but it is however fixed now by an explicit check.

### DIFF
--- a/mdp/utils/_symeig.py
+++ b/mdp/utils/_symeig.py
@@ -32,7 +32,7 @@ def _greatest_common_dtype(alist):
             dtype = _type_conv[transition]
     return dtype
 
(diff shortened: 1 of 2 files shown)


Stewori:
This fixes https://github.com/mdp-toolkit/mdp-toolkit/issues/32

PR #35, opened 2018-02-05
FIX: compatibility with Python 3

AlbertoEsc:
Removed calls to xrange and fixed some float divisions that should have been integer divisions


Stewori:
This looks good to me and seems to pass all checks. Thanks for committing! Accepting it right away...

PR #38, opened 2018-05-08
Added restructuredtext docstrings to all node modules

egbn17:
Dear MDP developers,

these commits contain the existing docstrings converted to the restructured text docstring format supported by the Sphinx as well as the Epydoc documentation generator, which are currently in use. Next to standardising the format, a completion of the descriptions was attempted where possible. Note that the commits limit themselves to the node modules contained in MDP. This should not prove to be a problem, as the parsing of rST using a non-standard format works out the way it has before. Testing has been done as per instructions on the MDP website.


mdp/nodes/classifier_nodes.py
```diff
@@ -85,18 +104,39 @@ def _train(self, x, labels):
             self.offset_weight = new_offset
 
     def _label(self, x):
-        """Returns an array with class labels from the perceptron.
+        """  
+        :param x: A matrix having different variables on different columns
+            and observations on the rows.
```
Stewori: why not add ``:type x: numpy.ndarray``?

mdp/nodes/classifier_nodes.py
```diff
@@ -209,12 +269,16 @@ def __init__(self, execute_method=None,
...
+        :rtype: list
+        """
         return ['b']
 
     def _train(self, x):
         """Provide the hopfield net with the possible states.
-
-        x -- a matrix having different variables on different columns
+        
+        :param x: A matrix having different variables on different columns
```
Stewori: why not add ``:type x: numpy.ndarray``?

mdp/nodes/classifier_nodes.py
```diff
@@ -228,24 +292,36 @@ def _train_one(self, pattern):
...
-        # we could use numx.fill_diagonal, but thats numpy 1.4 only
+        # we could use numx.fill_diagonal, but thats numpy 1. 4 only
         for i in range(self.input_dim):
             self._weight_matrix[i][i] = 0
 
     def _label(self, x, threshold = 0):
         """Retrieves patterns from the associative memory.
+        
+        :param x: A matrix having different variables on different columns
+            and observations on rows.
```
Stewori: why not add ``:type x: numpy.ndarray``?

mdp/nodes/expansion_nodes.py
```diff
@@ -299,8 +395,12 @@ class GeneralExpansionNode(_ExpansionNode):
 
     This node has been designed to facilitate nonlinear, fixed but arbitrary  
     transformations of the data samples within MDP flows.
-        
-    **Example**::
+    
+    Original code contributed by Alberto Escalante.
+    
+    |
```
Stewori: What is this?

mdp/nodes/expansion_nodes.py
```diff
@@ -358,14 +475,21 @@ def is_invertible():
...
+        This method requires scipy.
+        
+        :param x: The data.
+        :type x: numpy.ndarray
+        
+        :param use_hint: When calculating a pseudo inverse of the expansion,
+            the hint determines the starting point for the approximation.
+            For details on this parameter see the function 
+            ``invert_exp_funcs2`` in ``mdp.utils.routines.py``.
+        :type: numpy.ndarray or bool
```
Stewori: Should this rather be ``:type use_hint:`` ?

Stewori:
I skimmed through the first five files or so. First of all, this looks excellent to me, it really tidies up the doc. (I see that the plain sphinx autodoc style is the best solution w.r.t. compatibility although it's unfortunately best looking, but we had that discussion extensively...)
I added some cosmetical comments, e.g. from time to time a type specification seems to be missing. Please read through the diff view once again and fix these little things. Then, from my perspective we can merge it, if there are no concerns...

mdp/nodes/classifier_nodes.py
egbn17: IIRC I was not sure at the time whether MDP generalized to iterables. Due to limited knowledge of the specific modules, I was easy on said types. If that abstraction is not featured for this class, I will adapt the docstring. See http://mdp-toolkit.sourceforge.net/tutorial/iterables.html

mdp/nodes/classifier_nodes.py
egbn17: IIRC I was not sure at the time whether MDP generalized to iterables. Due to limited knowledge of the specific modules, I was easy on said types. If that abstraction is not featured for this class, I will adapt the docstring. See http://mdp-toolkit.sourceforge.net/tutorial/iterables.html

mdp/nodes/classifier_nodes.py
egbn17: IIRC I was not sure at the time whether MDP generalized to iterables. Due to limited knowledge of the specific modules, I was easy on said types. If that abstraction is not featured for this class, I will adapt the docstring. See http://mdp-toolkit.sourceforge.net/tutorial/iterables.html

mdp/nodes/expansion_nodes.py
egbn17: Do you mean Alberto or the vertical bar? The former contributed the code, the latter results in more line space between elements rendered on the website later on. I used those whenever it would result in a nicer display.

mdp/nodes/expansion_nodes.py
egbn17: adapted it

mdp/nodes/expansion_nodes.py
Stewori: I meant the vertical bar ;)
Is it somehow possible to add an explaining comment?
It easily looks like a typo and could be cleaned out by a future maintainer.

mdp/nodes/classifier_nodes.py
Stewori: Okay, I see. We should try to figure out the actual type in these cases.
Yes, often np.ndarray actually means 'arraylike' in the sense "everything that np.array eats without complaint" but thats (hopefully) a distinct issue.
What about ``numpy.ndarray or iterable of numpy.ndarray``?
Here PEP 484 notation might be favorable, i.e. ``Union[numpy.ndarray, Iterable[numpy.ndarray]]``. But then we should use it consistently; not a box I'd like to open at this stage of doc refactoring.

egbn17: As discussed, I have added the `numpy.ndarray` type and a `# todo: consider iterables` comment for each of the cases.

Issue #39, opened 2018-06-15
pytest 3.5: Metafunc.addcall is deprecated


Stewori:
Splitting off from https://github.com/mdp-toolkit/mdp-toolkit/pull/37. I did not yet investigate, just copied the issue description provided by @AlbertoEsc:

> pytest 3.5 gives hundreds of warnings “Metafunc.addcall is deprecated and scheduled to be removed in pytest 4.0. Please use Metafunc.parametrize instead.”, perhaps we should consider updating this part of the tests.

Stewori:
Was fixed by https://github.com/mdp-toolkit/mdp-toolkit/pull/40.

PR #40, opened 2018-06-23
swapped out deprecated metafunc.addcall method in pytest

NiMlr:
This should fix the [issue 39](https://github.com/mdp-toolkit/mdp-toolkit/issues/39).
The changes are based on this [part of the documentation](https://docs.pytest.org/en/3.0.5/parametrize.html). Test ran fine locally and deprecation warnings are gone, as expected.


Stewori:
Thanks for this work! The fix looks straight forward and tests pas, so merging it right away.

PR #43, opened 2018-11-02
Adding an expansion node for expansion with function sequences that can be recursively computed.

NiMlr:
Dear MDP developers,

I am happy to share the code of a new node to be considered for inclusion into MDP. The RecursiveExpansionNode implements recursion formulas for (orthogonal or linearly independend) sequences of functions, such as standard polynomials, Legendre polynomials, Legendre rational functions and Chebyshev polynomials of first kind. It can be extended easily to other recursions. Furthermore, an efficient way of computing multivariate expansions is employed.

In a following commit, a trainable version of this node is to be included. The Node was jointly developed by Stefan Richthofer and me.

### DIFF
--- a/mdp/nodes/recursive_expansion_nodes.py
+++ b/mdp/nodes/recursive_expansion_nodes.py
@@ -157,7 +157,7 @@ def init_legendre_poly(result, x, pos, cur_var):
     .. note::
 
         The procedure on how an init function is build can be found in the
-        docstring of the init method for
+        docstring of the init function for
         standard polynomials *init_standard_poly*.
 
 
@@ -468,7 +468,10 @@ class RecursiveExpansionNode(PolynomialExpansionNode):
 
         :param recf: Must be in ['standard_poly', 'legendre_poly',
             'legendre_rational', 'chebyshev_poly'] or a tuple similar
-            to those in the recfs dictionary in this module.
+            to those in the recfs dictionary in this module. The procedure
+            on how an init function is build can be found in the docstring
+            of the init and recursion function for standard polynomials 
+            *init_standard_poly* and *recf_standard_poly*, respectively.
         :type recf: tuple or str
 
         :param check: Indicates whether the input data will
--- a/mdp/nodes/recursive_expansion_nodes.py
+++ b/mdp/nodes/recursive_expansion_nodes.py
@@ -23,11 +23,18 @@ All orthogonal sequences of functions implemented require input values
 in the right scope, i.e. where the respective polynomials
 are designed to be orthogonal, usually values in [0, 1] or [-1, 1].
 That means, one might have to scale the input appropriately.
-The *TrainableRecursiveExpansionNode* is advisable to use, if the data supplied
+The *NormalizingRecursiveExpansionNode* is advisable to use, if the data supplied
 drops out of this interval or it corresponding cube (i.e. [0,1]^k).
-For convenience, the *TrainableRecursiveExpansionNode* can adapt the
+For convenience, the *NormalizingRecursiveExpansionNode* can adapt the
 value range of the data during a training phase and apply the corresponding
 scaling on execution.
+
+|
+
+.. admonition:: Reference
+
+    Example where orthogonal expansions are advisable to use:
+    URL: https://arxiv.org/abs/1805.08565, section 4.2.1, Figure 11 and A.2.
 """
 import mdp
 from mdp import numx as np
@@ -517,10 +524,10 @@ class RecursiveExpansionNode(_ExpansionNode):
             self.recf = recf[1]
             self.lower = recf[2]
             self.upper = recf[3]
-    
+
     def _get_supported_dtypes(self):
         """Return the list of dtypes supported by this node.
-        
+
         :return: The list of dtypes supported by this node.
         :rtype: list
         """
@@ -618,7 +625,7 @@ class RecursiveExpansionNode(_ExpansionNode):
                 "One or more values lie outside of the function specific domain.")
 
 
-class TrainableRecursiveExpansionNode(RecursiveExpansionNode):
+class NormalizingRecursiveExpansionNode(RecursiveExpansionNode):
     """Recursively computable (orthogonal) expansions and a
     trainable transformation to the domain of the expansions.
 
@@ -635,7 +642,7 @@ class TrainableRecursiveExpansionNode(RecursiveExpansionNode):
 
     def __init__(self, degree=1, recf='standard_poly', check=True, with0=True,
                  input_dim=None, dtype=None):
-        """Initialize a TrainableRecursiveExpansionNode.
+        """Initialize a NormalizingRecursiveExpansionNode.
 
         :param degree: The maximum order of the recursive expansion.
             The dimension of the return for single variable inputs will be
@@ -668,12 +675,12 @@ class TrainableRecursiveExpansionNode(RecursiveExpansionNode):
             Default is None.
         :type dtype: numpy.dtype or str
         """
-        super(TrainableRecursiveExpansionNode, self).__init__(degree,
-                                                              recf=recf,
-                                                              check=check,
-                                                              with0=with0, 
-                                                              input_dim=input_dim,
-                                                              dtype=dtype)
+        super(NormalizingRecursiveExpansionNode, self).__init__(degree,
+                                                                recf=recf,
+                                                                check=check,
+                                                                with0=with0,
+                                                                input_dim=input_dim,
+                                                                dtype=dtype)
 
         self.amaxcolumn = None
         self.amincolumn = None
@@ -688,7 +695,7 @@ class TrainableRecursiveExpansionNode(RecursiveExpansionNode):
         :type x: numpy.ndarray
         """
         self.domain_transformation(x)
-        return super(TrainableRecursiveExpansionNode, self)._execute(x)
+        return super(NormalizingRecursiveExpansionNode, self)._execute(x)
 
     def _train(self, x):
         """Determine coordinatewise and absolute maxima and minima.
--- a/mdp/nodes/recursive_expansion_nodes.py
+++ b/mdp/nodes/recursive_expansion_nodes.py
@@ -440,15 +440,22 @@ class RecursiveExpansionNode(PolynomialExpansionNode):
         result[:, 0] = 1.
         pos = 1
 
-        for cur_var in range(num_vars):
-            # preset index for current variable
-            pos, n, special = self.r_init(result, x, pos, cur_var)
-            # single variable recursion
-            while n <= deg:
-                        # recursion step
-                result[:, pos] = self.recf(result, x, special, n, cur_var, pos)
-                n += 1
-                pos += 1
+        if self._degree > 1:
+            for cur_var in range(num_vars):
+                # preset index for current variable
+                pos, n, special = self.r_init(result, x, pos, cur_var)
+                # single variable recursion
+                while n <= deg:
+                            # recursion step
+                    result[:, pos] = self.recf(
+                        result, x, special, n, cur_var, pos)
+                    n += 1
+                    pos += 1
+        # in case input is unreasonable
+        else:
+            result[:, 0] = 1
+            for i in range(num_vars):
+                result[:, i+1] = x[:, i]
 
         todoList = []
         for i in range(1, num_vars):
--- a/mdp/nodes/recursive_expansion_nodes.py
+++ b/mdp/nodes/recursive_expansion_nodes.py
@@ -6,9 +6,9 @@ The nodes contained in this module complement the existing
 computed sequences of basis function and by providing implemented
 polynomial and rational recursions that are numerically stable, even
 for high orders. More specifically this module provides the means
-for computing Legendre, Hermite or Chebyshev polynomials if first kind.
+for computing Legendre, Hermite or Chebyshev polynomials of first kind.
 
-Analytically all these polynomial expansions cover the same function space
+Analytically these polynomial expansions cover the same function space
 if they are of same degree. However the conventional
 monome-based polynomial expansion is numerically defective beyond
 degrees of about 20. Input values below one are flattened to
@@ -24,9 +24,9 @@ in the right scope, i.e. where the respective polynomials
 are designed to be orthogonal, usually values in [0, 1] or [-1, 1].
 That means, one might have to scale the input appropriately.
 The *TrainableRecursiveExpansionNode* is advisable to use, if the data supplied
-drops out of this interval or it corresponding cube (i.e. [0,1}^k).
+drops out of this interval or it corresponding cube (i.e. [0,1]^k).
 For convenience, the *TrainableRecursiveExpansionNode* can adapt the
-value range of the data during a training phase and applythe corresponding
+value range of the data during a training phase and apply the corresponding
 scaling on execution.
 """
 import mdp
--- a/mdp/nodes/recursive_expansion_nodes.py
+++ b/mdp/nodes/recursive_expansion_nodes.py
@@ -74,7 +74,7 @@ def init_standard_poly(result, x, pos, cur_var):
 
     :returns: The index of the next element that will be computed in the
         recursion, the order of the next element, and the index of a special
-        member.
+        element.
     :rtype: Tuple[int, int, int]
     """
     result[:, pos] = x[:, cur_var]
@@ -179,7 +179,8 @@ def init_legendre_poly(result, x, pos, cur_var):
 
     :returns: The index of the next element that will be computed in the
         recursion, the order of the next element, and the index of a special
-        member.
+        element. Is this case the first order Legendre polynomial of the 
+        current variable is the special element.
     :rtype: Tuple[int, int, int]
     """
     result[:, pos] = x[:, cur_var]
@@ -208,7 +209,8 @@ def recf_legendre_poly(result, x, special, n, cur_var, pos):
     :type x: numpy.ndarray
 
     :param special: Index of a special element to be considered in the
-        recursion formula. E.g. the first order function value.
+        recursion formula. In this case the special element is the 
+        first order Legendre polynomial of the current variable.
     :type special: int
 
     :param n: The order of the function to be computed at this step.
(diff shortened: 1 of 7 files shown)


Stewori:
Nils, thanks for setting this up. I will have a closer look soon.

Note to other readers as it might not be clear from the description above:

One might ask why additional expansion nodes besides the existing PolynomialExpansionNode are required, especially if they are also polynomial in essence like Legendre, Hermite or Chebyshev polynomials. Analytically all these polynomial expansions cover the same function space if they are of same degree. However the conventional monome-based polynomial expansion is numerically defective beyond degrees of about 20. Input values below one are flattened to zero and values above one explode. Analytically this would be compensated by very small or large coefficients that are not properly presentable in floating point. The orthogonal polynomials implemented here are evaluated using a numerically stable recursion formula. This only works for input values in the right scope, i.e. where the respective polynomials are designed to be orthogonal, usually values in [0, 1] or [-1, 1].
That means, one might have to scale the input appropriately. Here the trainable version of the node comes into play. For convenience, it can learn the value range of the data during a training phase and apply the corresponding scaling on execution.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+                 with0=True, input_dim=None, dtype=None):
+        """Initialize a RecursiveExpansionNode.
+
+        :param degree: The maximum order of the recursive expansion.
+            The dimension of the return for single variable inputs will be
+            equal to this value if with0 == False.
+        :type degree: int
+
+        :param recf: Must be in ['standard_poly', 'legendre_poly',
+            'legendre_rational', 'chebyshev_poly'] or a tuply similar
```
Stewori: tuply?

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+            to those in the recfs dictionary in this module.
```
Stewori: Drop a note here how the user would use his own recursion formula.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+    return (2.*n-1.)/n*(x[:, cur_var]-1.) / (x[:, cur_var]+1.) \
+        * result[:, pos-1] - (n-1.) / n * result[:, pos-2]
+
+
+def init_chebyshev_poly(result, x, pos, cur_var):
+    """Initialize the first and second order before starting the recursion.
+
+    Mathematically this amounts to
+    T_1 = x
+    T_2 = 2xT_1 -1.
```
Stewori: For better readablility write it 2x T_1 - 1 or even 2x * T_1 - 1. Omitting * when multi-character variable names are involved should be considered carefully.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+        member.
+    :rtype: numpy.ndarray
+    """
+    result[:, pos] = x[:, cur_var]
+    # pos, n, special
+    return pos+1, 2, pos
+
+
+def init_legendre_poly(result, x, pos, cur_var):
+    """Initialize  the first and second order before starting the recursion.
```
Stewori: Double space.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+
+    :returns: The index of the next element that will be computed in the
+        recursion, the order of the next element, and the index of a special
+        member.
+    :rtype: numpy.ndarray
+    """
+    result[:, pos] = x[:, cur_var]
+    # after the first variable this is needed (no zero order available)
+    result[:, pos+1] = 3./2.*result[:, pos] * result[:, pos] - \
+        1./2.
```
Stewori: Write .5 instead. Remove the linebreak.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
+"""
+This module provides nodes for recursive computation of sequences of functions.
+"""
+import mdp
+from mdp import numx as np
+from mdp.nodes import PolynomialExpansionNode
+__docformat__ = "restructuredtext en"
+
+
```
Stewori: Put the functions below into consistent order, i.e. always
init_x
recf_x

or always

recf_x
init_x.
Currently it's someimes init first, sometimes recf first.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+    :type int: int
+
+    :param cur_var: The index of the current variable to be considered in the
+        recursion. This value will have to be lower than x.shape[1].
+    :type cur_var: int
+
+    :returns: The index of the next element that will be computed in the
+        recursion, the order of the next element, and the index of a special
+        member.
+    :rtype: numpy.ndarray
```
Stewori: The return type is actually (int, int, int), or better as PEP 484 Tuple[int, int, int]?
Same in doc of the other init functions.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+
+        :param degree: The maximum order of the recursive expansion.
+            The dimension of the return for single variable inputs will be
+            equal to this value if with0 == False.
+        :type degree: int
+
+        :param recf: Must be in ['standard_poly', 'legendre_poly',
+            'legendre_rational', 'chebyshev_poly'] or a tuply similar
+            to those in the recfs dictionary in this module.
+        :type recf: tuple
```
Stewori: tuple or string

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+        """
+        xmax = np.amax(x)-prec
+        xmin = np.amin(x)+prec
+
+        if (self.upper < xmax) or (self.lower > xmin):
+            raise mdp.NodeException(
+                "One or more values lie outside of the function specific domain.")
+
+    @staticmethod
+    def is_trainable():
```
Stewori: Is this function needed? Insn't False the default already?

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+        return False
+
+    @staticmethod
+    def is_invertible():
+        return False
+
+
+class TrainableRecursiveExpansionNode(RecursiveExpansionNode):
```
Stewori: I wonder if we can somehow shorten this name.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+            if (xmax < self.upper) and (xmin > self.lower):
+                def f(x):
+                    return x
+            elif datamaxdev < dev:
+                def f(x):
+                    x += (-datamean)+mean
+                    return x
+            else:
+                def f(x):
+                    x += ((-datamean)+mean*datamaxdev/dev)
```
Stewori: outer brackets are unnecessary

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+        if self.amin is None:
+            self.amaxcolumn = np.amax(x, axis=0)
+            self.amincolumn = np.amin(x, axis=0)
+            self.amax = np.amax(x)
+            self.amin = np.amin(x)
+        else:
+            self.amaxcolumn = np.maximum(self.amaxcolumn, np.amax(x, axis=0))
+            self.amincolumn = np.minimum(self.amincolumn, np.amin(x, axis=0))
+            self.amax = np.maximum(self.amax, np.amax(x))
+            self.amin = np.minimum(self.amin, np.amin(x))
```
Stewori: I don't really see why we need amaxcolumn / amincolumn. Considering that data can be high-dimensional, this could waste memory.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,623 @@
...
+            must be along the first axis, different variables along the second.
+        :type x: numpy.ndarray
+
+        The values are used to generate a transformation to the valid domain
+        for the data.
+        """
+        if self.amin is None:
+            self.amaxcolumn = np.amax(x, axis=0)
+            self.amincolumn = np.amin(x, axis=0)
+            self.amax = np.amax(x)
```
Stewori: Can't this be computed as np.max(self.amaxcolumn)? No need to search the whole chunk twice. Searching amaxcolumn is much cheaper. Same for min.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: fixed

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Fixed.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Fixed.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Fixed.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Fixed.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Fixed.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Fixed.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: It is. I omitted the methods in the new version.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Fixed.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Fixed.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Fixed. Thanks for the review!

mdp/nodes/recursive_expansion_nodes.py
Stewori: Is it then not sufficient to translate all data by amax / amin? If you translate by amaxcolumn, eachcomponent is translated differently, right? Wouldn't that cause a translational shift between the components? I doubt that this would be desired.

Stewori:
A polished variant of the explanation I gave in https://github.com/mdp-toolkit/mdp-toolkit/pull/43#issuecomment-444059846 should go into the module top-level docstring.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Done

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+        expansion of the current variable, namely::
+
+            >>> result[:, pos]
+
+        We access all observations of the n-1-th order expansion of the current
+        variable at::
+
+            >>> result[:, pos-1]
+
+        Using the the index of the special variable we specified in our init
```
Stewori: double 'the'

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+
+            >>> result[:, pos-1]*result[:, special]
+
+        Alternatively, in this case we could have used the x variable directly,
+        by setting the return value to::
+
+            >> result[:, pos-1]*x[:, cur_var]
+
+        The recursion is natively applied to all variables contained in the data
+        one after another and the results saved in the
```
Stewori: replace 'saved' with 'are stored'

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+        one-dimensional result array.
+
+
+    :param result: Contains the observations along the first dimension
+        and the different function values of expansion w.r.t. an observation
```
Stewori: remove 'different'

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+def init_legendre_poly(result, x, pos, cur_var):
+    """Initialize the first and second order before starting the recursion.
+
+    Mathematically this amounts to
+    P_1 = x
+    P_2 = 3/2 * P_1*P_1 - 1/2 .
+
+    .. note::
+
+        The procedure on how an init function is build can be found in the
```
Stewori: is build -> is built
https://www.dictionary.com/browse/built

mdp/nodes/recursive_expansion_nodes.py
Stewori: (also below, wherever it occurs like this.)

mdp/nodes/recursive_expansion_nodes.py
Stewori: (also change in further occurrences of "is build" below)

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+    return pos
+
+
+class RecursiveExpansionNode(PolynomialExpansionNode):
+    """Recursively computable (orthogonal) expansions.
+
+    .. attribute:: lower
+
+        The lower bound of the domain, on which the recursion
+        function sequence selected, is defined or orthogonal.
```
Stewori: The commas here are wrong, because "on which the recursion function sequence selected" is defining and the other clause too. However I'd simplify the whole thing by just remove 'sequence selected':

The lower bound of the domain on which the recursion function is defined or orthogonal.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+
+    .. attribute:: upper
+
+        The upper bound of the domain, on which the recursion
+        function sequence selected, is defined or orthogonal.
```
Stewori: like above

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+    """
+
+    def __init__(self, degree=1, recf='standard_poly', check=False,
+                 with0=True, input_dim=None, dtype=None):
```
Stewori: I still wonder if we can find a better name than 'with0'.

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+    :param result: Contains the observations along the first dimension
+        and the function values of expansion w.r.t. an observation
+        along the second dimension.
+    :type result: numpy.ndarray
+
+    :param x: The data to be expanded.
+    :type x: numpy.ndarray
+
+    :param special: Index of a special element to be considered in the
+        recursion formula. E.g. the first order function value.
```
Stewori: Please replace the generic explanation here and in all other recf functions except the standard one. Describe the specific role of ``special`` for Legendre polynomials etc...

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+    .. attribute:: lower
+
+        The lower bound of the domain on which the recursion function is
+        defined or orthogonal.
+
+    .. attribute:: upper
+
+        The lower bound of the domain on which the recursion function is
+        defined or orthogonal.
+    """
```
Stewori: Can we add a note on why this extends PolynomialExpansionNode? I.e. because the tensor basis expansion (w.r.t. simplex form) has similar expansion dim behavior. This gives us convenient access to ``expanded_dim``. What about ``_get_supported_dtypes``? Int and float seem okay to me, except that LegendreRational functions perform some division. Does this work through integer arithmetics? In Python 3 ``/`` yields a float result. Maybe we should exclude Integer support for this. Maybe -- for simplicity -- we should exclude integer support at all.

It seems to me that we might better extend ``_ExpansionNode`` rather than ``PolynomialExpansionNode``. What do you think. Then we'd have to access the static ``expanded_dim`` like here: https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/nodes/expansion_nodes.py#L114

mdp/nodes/recursive_expansion_nodes.py
```diff
@@ -0,0 +1,767 @@
...
+    """Recursively computable (orthogonal) expansions.
+
+    .. attribute:: lower
+
+        The lower bound of the domain on which the recursion function is
+        defined or orthogonal.
+
+    .. attribute:: upper
+
+        The lower bound of the domain on which the recursion function is
```
Stewori: lower -> upper

Stewori: please check if this c/p error w.r.t lower/upper occurs elsewhere

NiMlr: Done.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: Done.

mdp/nodes/recursive_expansion_nodes.py
NiMlr: It seems more reasonable to me to not extend PolynomialExpansionNode. We are now extending _ExpansionNode and using the expanded_dim function provided in the ExpansionNodes module. Furthermore, I restricted the supported datatypes to 'AllFloat', as I am losing interpretation for a lot of these function sequences on integer values. Please let me know if you have a reason to include interger datatypes I have not thought of.

Stewori:
I have opened #44 for the Travis CI crash.

Stewori:
I am fine with accepting this now. However I will likely edit the documentation at some places in a subsequent commit. That's just about details in some formulations.
Thanks for this work and good job!

PR #45, opened 2019-01-25
FIX: Added some cornercase handling

NiMlr:
Zero order expansions lead to an error in the current implementation. So I am providing this fix.


Stewori:
I should have merged this earlier. Sorry for the inconvenience.

Issue #36, opened 2018-02-08
Letting Travis CI test on Python 3 as well


Stewori:
Following up on @AlbertoEsc's latest fix, we should let Travis test on Python 3 as well, given that all tests pass now on Python 3. Looking at [.travis.yml](https://github.com/mdp-toolkit/mdp-toolkit/blob/master/.travis.yml) I recognize that mdp uses a hack to let Travis access numpy and scipy from neurodebian. I'm not familiar with this procedure, usually I would have something like this in travis.yml:

```
python:
  - "2.7"
  - "3.3"
  - "3.4"
  - "3.5.2"
  - "3.5.3"
- "3.6
```
Does someone know how to achieve Python 3 tests based on the current hack (i.e. pretending to be erlang)? @otizonaizit ?
I confirmed that with current config, Travis really does only run tests on Python 2.7.

otizonaizit:
the hack is a relic of the past, when installing numpy and/or scipy on travis meant having to compile fortran code and the whole procedure would take more than 30 minutes to complete. 

Nowadays with wheels installing latest numpy and scipy with pip is very fast, so the whole travis.yml can be rewritten in a much more standard form. If you do it, maybe add a couple of older releases of scipy and numpy into the matrix, so that you check that you don't break backward compatibility by mistake…

If I am not mistaken, I think that we don't even need a virtualenv in travis, a `pip install -e` may be enough for testing against the latest releases of all dependencies…

Stewori:
Fixed as of b711200bf27fceb33d19e71c8c7be4eeeb2f59e8

PR #46, opened 2019-02-22
FIX: .travis.yml to test different python versions

NiMlr:
Addressing [issue 36](https://github.com/mdp-toolkit/mdp-toolkit/issues/36) I have prepared this fix. It should now be possible to automatically run tests on all python versions that will be supported longer than mid March 2019. Tests on multiple numpy versions have been incorporated as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,8 @@ matrix:
       env: VNUMPY=1.8.2
     - python: "3.6"
       env: VNUMPY=1.8.2
+    - python: "3.5-dev"
+      env: VNUMPY=1.8.2
       
 before_install:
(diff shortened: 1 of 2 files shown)


NiMlr:
There seems to be a precision problem in test_ccipcanode_v2.

NiMlr:
After easing the precision in test_ccipcanode_v2 the tests pass.

Stewori:
Thanks for this work! This looks good to me. Switching to a pip-supplied numpy is the only way to test against a wide range of numpy versions, especialy newer ones, right? So, I am fine with that decision and since nobody else spoke up I'll merge this now.

PR #48, opened 2019-06-30
SFA with non constant time increments

NiMlr:
Dear MDP developers,

I am happy to propose this generalization of SFA, that works with any strictly positive time increments of signal values.
The main adaptions include:

- A derivative approximation, that takes the step size into account (equal to the original SFA with step size 1).
- Generalization of the covariance estimation to be an approximation of the pairwise scalar product of continuous functions with one variable and new implementation of the approximation using the trapezoidal rule. This is only used to estimate the covariance matrix for whitening and the derivative signal in SFA.

Tests have been implemented - among some comparing this generalization to the original SFA implementation.

### DIFF
--- a/mdp/nodes/sfa_nodes.py
+++ b/mdp/nodes/sfa_nodes.py
@@ -376,7 +376,7 @@ class SFANode(Node):
         return self._refcast(t / (2 * numx.pi) * numx.sqrt(self.d))
 
 
-class UnevenlySampledSFANode(Node):
+class UnevenlySampledSFANode(SFANode):
     """
     Extract the slowly varying components from the input data.
     This node can be understood as a generalization to the *SFANode*.
@@ -455,23 +455,9 @@ class UnevenlySampledSFANode(Node):
             ``train()`` calls.
         :type rank_deficit_method: str
         """
-        super(UnevenlySampledSFANode, self).__init__(
-            input_dim, output_dim, dtype)
-
-        # init two covariance matrices
-        self._init_cov()
-
-        # set routine for eigenproblem
-        self.set_rank_deficit_method(rank_deficit_method)
-        self.rank_threshold = 1e-12
-        self.rank_deficit = 0
-
-        # SFA eigenvalues and eigenvectors, will be set after training
-        self.d = None
-        self.sf = None  # second index for outputs
-        self.avg = None
-        self._bias = None  # avg multiplied with sf
-        self.tlen = None
+        super(UnevenlySampledSFANode, self).__init__(input_dim,
+                                                     output_dim, dtype, include_last_sample=True,
+                                                     rank_deficit_method=rank_deficit_method)
 
         self.tphase = 0
 
@@ -552,168 +538,6 @@ class UnevenlySampledSFANode(Node):
         else:
             dt_ = None
         self._dcov_mtx.update(x_, dt_)
-
-    def set_rank_deficit_method(self, rank_deficit_method):
-        if rank_deficit_method == 'pca':
-            self._symeig = symeig_semidefinite_pca
-        elif rank_deficit_method == 'reg':
-            self._symeig = symeig_semidefinite_reg
-        elif rank_deficit_method == 'svd':
-            self._symeig = symeig_semidefinite_svd
-        elif rank_deficit_method == 'ldl':
-            try:
-                from scipy.linalg.lapack import dsytrf
-            except ImportError:
-                err_msg = ("ldl method for solving SFA with rank deficit covariance "
-                           "requires at least SciPy 1.0.")
-                raise NodeException(err_msg)
-            self._symeig = symeig_semidefinite_ldl
-        elif rank_deficit_method == 'auto':
-            self._symeig = symeig_semidefinite_pca
-        elif rank_deficit_method == 'none':
-            self._symeig = symeig
-        else:
-            raise ValueError("Invalid value for rank_deficit_method: %s"
-                             % str(rank_deficit_method))
-
-    def _set_range(self):
-        if self.output_dim is not None and self.output_dim <= self.input_dim:
-            # (eigenvalues sorted in ascending order)
-            rng = (1, self.output_dim)
-        else:
-            # otherwise, keep all output components
-            rng = None
-            self.output_dim = self.input_dim
-        return rng
-
-    def _check_train_args(self, x, *args, **kwargs):
-        """
-        Raises exception if time dimension does not have enough elements.
-
-        :param x: The time series data.
-        :type x: numpy.ndarray
-
-        :param *args:
-        :param **kwargs:
-        """
-        # check that we have at least 2 time samples to
-        # compute the update for the derivative covariance matrix
-        s = x.shape[0]
-        if s < 2:
-            raise TrainingException('Need at least 2 time samples to '
-                                    'compute time derivative (%d given)' % s)
-
-    def _stop_training(self, debug=False):
-        # request the covariance matrices and clean up
-        if hasattr(self, '_dcov_mtx'):
-            self.cov_mtx, self.avg, self.tlen = self._cov_mtx.fix()
-            del self._cov_mtx
-        # do not center around the mean:
-        # we want the second moment matrix (centered about 0) and
-        # not the second central moment matrix (centered about the mean), i.e.
-        # the covariance matrix
-        if hasattr(self, '_dcov_mtx'):
-            self.dcov_mtx, self.davg, self.dtlen = self._dcov_mtx.fix(
-                center=False)
-            del self._dcov_mtx
-
-        rng = self._set_range()
-
-        # solve the generalized eigenvalue problem
-        # the eigenvalues are already ordered in ascending order
-        try:
-            try:
-                # We first try to fulfill the extended signature described
-                # in mdp.utils.symeig_semidefinite
-                self.d, self.sf = self._symeig(
-                    self.dcov_mtx, self.cov_mtx, True, "on", rng,
-                    overwrite=(not debug),
-                    rank_threshold=self.rank_threshold, dfc_out=self)
-            except TypeError:
-                self.d, self.sf = self._symeig(
-                    self.dcov_mtx, self.cov_mtx, True, "on", rng,
-                    overwrite=(not debug))
-            d = self.d
-            # check that we get only *positive* eigenvalues
-            if d.min() < 0:
-                err_msg = ("Got negative eigenvalues: %s.\n"
-                           "You may either set output_dim to be smaller,\n"
-                           "or prepend the SFANode with a PCANode(reduce=True)\n"
-                           "or PCANode(svd=True)\n"
-                           "or set a rank deficit method, e.g.\n"
-                           "create the SFA node with rank_deficit_method='auto'\n"
-                           "and try higher values for rank_threshold, e.g. try\n"
-                           "your_node.rank_threshold = 1e-10, 1e-8, 1e-6, ..." % str(d))
-                raise NodeException(err_msg)
-        except SymeigException as exception:
-            errstr = (str(exception)+"\n Covariance matrices may be singular.\n"
-                      + SINGULAR_VALUE_MSG)
-            raise NodeException(errstr)
-
-        if not debug:
-            # delete covariance matrix if no exception occurred
-            del self.cov_mtx
-            del self.dcov_mtx
-
-        # store bias
-        self._bias = mult(self.avg, self.sf)
-
-    def _execute(self, x, n=None):
-        """
-        Compute the output of the slowest functions.
-        If 'n' is an integer, then use the first 'n' slowest components.
-
-        :param x: The time series data.
-        :type x: numpy.ndarray
-
-        :param n: The number of slowest components.
-        :type n: int
-
-        :returns: The output of the slowest functions.
-        :rtype: numpy.ndarray
-        """
-        if n:
-            sf = self.sf[:, :n]
-            bias = self._bias[:n]
-        else:
-            sf = self.sf
-            bias = self._bias
-        return mult(x, sf) - bias
-
-    def _inverse(self, y):
-        return mult(y, pinv(self.sf)) + self.avg
-
-    def get_eta_values(self, t=1):
-        """
-        Return the eta values of the slow components learned during
-        the training phase. If the training phase has not been completed
-        yet, call `stop_training`.
-
-        The delta value of a signal is a measure of its temporal
-        variation, and is defined as the mean of the derivative squared,
-        i.e. delta(x) = mean(dx/dt(t)^2).  delta(x) is zero if
-        x is a constant signal, and increases if the temporal variation
-        of the signal is bigger.
-
-        The eta value is a more intuitive measure of temporal variation,
-        defined as
-        eta(x) = t/(2*pi) * sqrt(delta(x))
-        If x is a signal of length 't' which consists of a sine function
-        that accomplishes exactly N oscillations, then eta(x)=N.
-
-        :param t: Sampling frequency in Hz.
-
-            The original definition in (Wiskott and Sejnowski, 2002)
-            is obtained for t = number of training data points, while
-            for t=1 (default), this corresponds to the beta-value defined
-            in (Berkes and Wiskott, 2005).
-
-        :returns: The eta values of the slow components learned during
-            the training phase.
-        """
-        if self.is_training():
-            self.stop_training()
-        return self._refcast(t / (2 * numx.pi) * numx.sqrt(self.d))
 
 
 class SFA2Node(SFANode):
--- a/mdp/nodes/sfa_nodes.py
+++ b/mdp/nodes/sfa_nodes.py
@@ -9,6 +9,11 @@ from mdp.utils import (mult, pinv, CovarianceMatrix,
                        symeig, SymeigException, symeig_semidefinite_reg,
                        symeig_semidefinite_pca, symeig_semidefinite_svd,
                        symeig_semidefinite_ldl)
+import warnings
+
+_INC_ARG_WARNING1 = ('As time_dependence is not specified, argument dt '
+                     'should be of length x.shape[0]-1.')
+warnings.filterwarnings('always', _INC_ARG_WARNING1, mdp.MDPWarning)
 
 SINGULAR_VALUE_MSG = '''
 This usually happens if there are redundancies in the (expanded) training data.
@@ -461,7 +466,7 @@ class VartimeSFANode(SFANode):
         # one for the derivatives
         self._dcov_mtx = VartimeCovarianceMatrix(self.dtype)
 
-    def time_derivative(self, x, dt=None):
+    def time_derivative(self, x, dt=None, time_dep=True):
         """Compute the linear approximation of the time derivative
 
         :param x: The time series data.
@@ -470,7 +475,11 @@ class VartimeSFANode(SFANode):
         :param dt: Sequence of time increments between vectors. 
 
             Usage with only single chunk of data:
-                *dt* must be of length *x.shape[0]-1*.
+                *dt* should be of length *x.shape[0]-1* or a constant. When
+                constant, the time increments are assumed to be constant. If
+                *dt* is not supplied, a constant time increments of one is
+                assumed. If a time sequence of length *x.shape[0]* is supplied
+                the first element is disregarded and a warning is presented.
 
             Usage with multiple chunks of data with intended time dependence:
                 *dt* should have length *x.shape[0]-1* in the first call and
@@ -478,44 +487,72 @@ class VartimeSFANode(SFANode):
                 Starting with the second call, the first element in each chunk
                 will be considered the time difference between the last element
                 of *x* in the previous chunk and the first element of *x* of the
-                current chunk.
+                current chunk. The *time_dep* argument should be *True*.
 
             Usage with multiple chunks without time dependence:
-                If *dt* has length *x.shape[0]-1* time dependence between
-                chunks is omitted and all algorithmic components regard
-                only the time structure within chunks.
+                The moments are computed as a weighted average of the moments
+                of separate chunks, if *dt* continues to have length
+                *x.shape[0]-1* after the first call. Time dependence between
+                chunks is thus omitted and all algorithmic components regard
+                only the time structure within chunks. The *time_dep* argument
+                should be *False*. As in the single chunk case it is possible
+                to set *dt* to be a constant or omit it completely for constant
+                or unit time increments within chunks.
 
-            Minimal usage:
-                If *dt* is omitted entirely there will be no time dependence
-                between chunks and it will be considered to be one everywhere
-                within a chunk.
+        :type dt: numpy.ndarray or numeric
 
-        :type dt: numpy.ndarray
+        :param time_dep: Indicates whether time dependence between chunks can
+            be considered. The argument is only relevant in case multiple chunks
+            of data are used. Time dependence between chunks is disregarded
+            when data collection has been done time independently and thus no
+            reasonable time increment between the end of a chunk and
+            beginning of the next can be specified.
+        
+        :type time_dep: bool
 
         :returns: Piecewise linear approximation of the time derivative.
         :rtype: numpy.ndarray
         """
-        outlen = x.shape[0] - 1\
-                if self.tchunk == 0 or dt is None or dt.shape[0] != x.shape[0] else x.shape[0]
+        if dt is not None and type(dt) == numx.ndarray:
+            # check for inconsistent arguments
+            if time_dep and self.tchunk >0 and x.shape[0] == len(dt)-1:
+                raise Exception('As time_dependence is specified, and it is not the first'
+                        '\ncall argument dt should be of length x.shape[0].')
+            if not time_dep and x.shape[0] == len(dt):
+                warnings.warn(_INC_ARG_WARNING1, mdp.MDPWarning)
+            if len(dt) != x.shape[0] and len(dt) != x.shape[0]-1:
+                raise Exception('Unexpected length of dt.')
+        elif dt is not None and not dt > 0:
+            raise Exception('Unexpected type or value of dt.')
+
+        if dt is None or type(dt) == numx.ndarray:
+            outlen = x.shape[0] - 1\
+                    if self.tchunk == 0 or dt is None or dt.shape[0] != x.shape[0] else x.shape[0]
+        elif dt > 0:
+            outlen = x.shape[0] - 1
         out = numx.empty([outlen, x.shape[1]])
 
-        if dt is not None:
+        if dt is not None and type(dt)==numx.ndarray:
             # Improvements can be made, by interpolating polynomials
             out[-x.shape[0]+1:, :] =\
                     (x[1:, :]-x[:-1, :]) / dt[-x.shape[0]+1:, None]
+        elif dt > 0:
+            out[-x.shape[0]+1:, :] = (x[1:, :]-x[:-1, :])/dt
         else:
             # trivial fallback
             out[-x.shape[0]+1:, :] = x[1:, :]-x[:-1, :]
 
-        if self.tchunk > 0 and dt is not None and dt.shape[0] == x.shape[0]:
+        if self.tchunk > 0 and time_dep and type(dt) == numx.ndarray and dt.shape[0] == x.shape[0]:
             out[0, :] = numx.divide(x[0, :]-self.xlast, dt[0])
+        elif self.tchunk > 0 and  time_dep and dt > 0:
+            out[0, :] = numx.divide(x[0, :]-self.xlast, dt)
 
         self.xlast = x[-1, :].copy()
         self.tchunk += 1
 
         return out
 
-    def _train(self, x, dt=None):
+    def _train(self, x, dt=None, time_dep=True):
         """Training method.
 
         :param x: The time series data.
@@ -524,7 +561,11 @@ class VartimeSFANode(SFANode):
         :param dt: Sequence of time increments between vectors. 
 
             Usage with only single chunk of data:
-                *dt* must be of length *x.shape[0]-1*.
+                *dt* should be of length *x.shape[0]-1* or a constant. When
+                constant, the time increments are assumed to be constant. If
+                *dt* is not supplied, a constant time increments of one is
+                assumed. If a time sequence of length *x.shape[0]* is supplied
+                the first element is disregarded and a warning is presented.
 
             Usage with multiple chunks of data with intended time dependence:
                 *dt* should have length *x.shape[0]-1* in the first call and
@@ -532,31 +573,42 @@ class VartimeSFANode(SFANode):
                 Starting with the second call, the first element in each chunk
                 will be considered the time difference between the last element
                 of *x* in the previous chunk and the first element of *x* of the
-                current chunk.
+                current chunk. The *time_dep* argument should be *True*.
 
             Usage with multiple chunks without time dependence:
-                If *dt* has length *x.shape[0]-1* time dependence between
-                chunks is omitted and all algorithmic components regard
-                only the time structure within chunks.
+                The moments are computed as a weighted average of the moments
+                of separate chunks, if *dt* continues to have length
+                *x.shape[0]-1* after the first call. Time dependence between
+                chunks is thus omitted and all algorithmic components regard
+                only the time structure within chunks. The *time_dep* argument
+                should be *False*. As in the single chunk case it is possible
+                to set *dt* to be a constant or omit it completely for constant
+                or unit time increments within chunks.
 
-            Minimal usage:
-                If *dt* is omitted entirely there will be no time dependence
-                between chunks and it will be considered to be one everywhere
-                within a chunk.
+        :type dt: numpy.ndarray or numeric
 
-        :type dt: numpy.ndarray
+        :param time_dep: Indicates whether time dependence between chunks can
+            be considered. The argument is only relevant in case multiple chunks
+            of data are used. Time dependence between chunks is disregarded
+            when data collection has been done time independently and thus no
+            reasonable time increment between the end of a chunk and
+            beginning of the next can be specified.
+        
+        :type time_dep: bool
         """
         # update the covariance matrices
-        self._cov_mtx.update(x, dt)
-        x_ = self.time_derivative(x, dt)
+        self._cov_mtx.update(x, dt, time_dep)
+        x_ = self.time_derivative(x, dt, time_dep)
 
         if (x_.shape[0] > x.shape[0]-1) and (dt is not None):
             dt_ = dt
-        elif dt is not None:
+        elif dt is not None and type(dt)==numx.ndarray:
             dt_ = dt[1:]
+        elif dt is not None and dt > 0:
+            dt_ = dt
         else:
             dt_ = None
-        self._dcov_mtx.update(x_, dt_)
+        self._dcov_mtx.update(x_, dt_, time_dep)
 
 
 class SFA2Node(SFANode):
--- a/mdp/nodes/sfa_nodes.py
+++ b/mdp/nodes/sfa_nodes.py
@@ -536,7 +536,7 @@ class VartimeSFANode(SFANode):
             # Improvements can be made, by interpolating polynomials
             out[-x.shape[0]+1:, :] =\
                     (x[1:, :]-x[:-1, :]) / dt[-x.shape[0]+1:, None]
-        elif dt > 0:
+        elif dt is not None and dt > 0:
             out[-x.shape[0]+1:, :] = (x[1:, :]-x[:-1, :])/dt
         else:
             # trivial fallback
--- a/mdp/nodes/sfa_nodes.py
+++ b/mdp/nodes/sfa_nodes.py
@@ -5,7 +5,7 @@ __docformat__ = "restructuredtext en"
 import mdp
 from mdp import numx, Node, NodeException, TrainingException
 from mdp.utils import (mult, pinv, CovarianceMatrix,
-                       UnevenlySampledCovarianceMatrix, QuadraticForm,
+                       VartimeCovarianceMatrix, QuadraticForm,
                        symeig, SymeigException, symeig_semidefinite_reg,
                        symeig_semidefinite_pca, symeig_semidefinite_svd,
                        symeig_semidefinite_ldl)
@@ -376,7 +376,7 @@ class SFANode(Node):
         return self._refcast(t / (2 * numx.pi) * numx.sqrt(self.d))
 
 
-class UnevenlySampledSFANode(SFANode):
+class VartimeSFANode(SFANode):
     """
     Extract the slowly varying components from the input data.
     This node can be understood as a generalization to the *SFANode*.
@@ -408,7 +408,7 @@ class UnevenlySampledSFANode(SFANode):
     def __init__(self, input_dim=None, output_dim=None, dtype=None,
                  rank_deficit_method='none'):
         """"
-        Initialize an object of type 'UnevenlySampledSFANode'.
+        Initialize an object of type 'VartimeSFANode'.
 
         :param input_dim: The input dimensionality.
         :type input_dim: int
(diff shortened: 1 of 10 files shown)


mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +371,66 @@ def get_eta_values(self, t=1):
         return self._refcast(t / (2 * numx.pi) * numx.sqrt(self.d))
 
 
+class UnevenlySampledSFANode(SFANode):
```
Stewori: In the docstring, point out the usecase of using SFA on data with missing values. I think missing data is really the main usecase of this node. Also point out that the linear interpolation here is only the first step and to serve as bottom line for testing potential future versions that feature better interpolation ideas (there is kind of a research field around estimating covariance with missing data).

mdp/test/test_Unevenly_Sampled_Covariance.py
```diff
@@ -0,0 +1,57 @@
...
+
+    assert_array_almost_equal(unC, tC, decimal-5)
+    assert_array_almost_equal(unC2, tC, decimal-5)
+
+    prec = numx.linalg.norm(tC-C)
+    unPrec = numx.linalg.norm(tC-unC)
+    unPrec2 = numx.linalg.norm(tC-unC2)
+
+    assert_array_almost_equal(prec, unPrec, decimal-3)
+    assert_array_almost_equal(prec, unPrec2, decimal-3)
```
Stewori: ``-3`` seems to me like a pretty low precision. I would expect it to be at about ``-10`` or at least ``-6`` or so. Is this value kind of sharp for the test results?

mdp/test/test_UnevenlySampledSFANode.py
```diff
@@ -0,0 +1,28 @@
...
+
+    # stop training and generate slow features
+    sfa.stop_training()
+    unsfa.stop_training()
+
+    if ((sfa.sf[0, 0] > 0) and (unsfa.sf[0, 0] < 0)) \
+            or ((sfa.sf[0, 0] < 0) and (unsfa.sf[0, 0] > 0)):
+        sfa.sf *= -1.
+
+    assert_array_almost_equal(unsfa.sf, sfa.sf, decimal-6)
```
Stewori: Another good test for evenly against unevenly sampled SFA occurred to me. Create some data, then define some of its values to be missing (e.g. at a number of random locations). Once use [numpy.interp](https://docs.scipy.org/doc/numpy/reference/generated/numpy.interp.html) to fill the wholes, on the other hand create an appropriate dt to fill the wholes. Then original SFA should yield the same result on the interpolated data like uneven SFA would yield on the original data with proper `dt`.

Stewori:
As a note for other readers:
A main usecase of this node is to provide a way to deal with missing data in SFA. Missing data is a common issue blocking the use of SFA on real-world data. Actually there are better ways to treat missing data than the linear interpolation applied here. However it is crucial as a bottom line for testing potential future versions. The whole thing boils down to estimating covariance matrices in context of missing data. There is kind of a research field around this, e.g. using Lomp-Scargle periodograms. Having a way to estimate covariance matrices in context of missing data can benefit several other covariance-based MDP-nodes as well, e.g. PCA.

mdp/nodes/sfa_nodes.py
```diff
@@ -290,7 +292,7 @@ def _stop_training(self, debug=False):
             # check that we get only *positive* eigenvalues
             if d.min() < 0:
                 err_msg = ("Got negative eigenvalues: %s.\n"
-                           "You may either set output_dim to be smaller,\n"
+                           "You may either set oucovartput_dim to be smaller,\n"
```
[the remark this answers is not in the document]
NiMlr: Done.

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +371,66 @@ def get_eta_values(self, t=1):
         return self._refcast(t / (2 * numx.pi) * numx.sqrt(self.d))
 
 
+class UnevenlySampledSFANode(SFANode):
+
+    def __init__(self, input_dim=None, output_dim=None, dtype=None,
+                 include_last_sample=True, rank_deficit_method='none'):
+        super(SFANode, self).__init__(input_dim, output_dim, dtype)
```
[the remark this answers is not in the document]
NiMlr: Thanks for the great suggestion!

Should `super(UnevenlySampledSFANode, self).__init__(input_dim, output_dim, dtype, include_last_sample, rank_deficit_method)`
not be sufficient?

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +371,66 @@ def get_eta_values(self, t=1):
...
+        Compute the linear approximation of the time derivative
+
+        :param x: The time series data.
+        :type x: numpy.ndarray
+
+        :returns: Piecewise linear approximation of the time derivative.
+        :rtype: numpy.ndarray
+        """
+        # Improvements can be made, by interpolating polynomials
+        return (x[1:, :]-x[:-1, :])/dt[:,None]
```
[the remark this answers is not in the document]
NiMlr: Done.

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +371,66 @@ def get_eta_values(self, t=1):
...
+    
+    def _train(self, x, dt, include_last_sample=None):
```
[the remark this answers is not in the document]
NiMlr: As `dt` is not optional, do you want to include an error raise? Otherwise we could set `dt= numx.ones(..)`.

mdp/utils/covariance.py
```diff
@@ -133,6 +135,120 @@ def fix(self, center=True):
...
+            self._init_internals(x)
+        # cast input
+        x = mdp.utils.refcast(x, self._dtype)
+        # update the covariance matrix, the average and the number of
+        # observations
+        # the implementation is analogous to the evenly sampled case
+        # values might get big as it is only normalized in fix
+        xcpy = numx.multiply(x[1:, :], numx.sqrt(dt/2.)[:, None])
+        self._cov_mtx += mdp.utils.mult(xcpy.T, xcpy)
+        xcpy[:, :] = numx.multiply(x[:-1, :], numx.sqrt(dt/2.)[:, None])
```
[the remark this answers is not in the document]
NiMlr: Oh it seems I was imprecise there. I have adapted it.

mdp/utils/covariance.py
```diff
@@ -133,6 +135,120 @@ def fix(self, center=True):
...
+            Default is None. If dtype is not defined, it will be inherited
+            from the first data bunch received by 'update'. 
+        :type dtype: numpy.dtype or str
+        """
+        super(UnevenlySampledCovarianceMatrix,
+              self).__init__(dtype=dtype, bias=True)
+        # we need to keep steps and actual time
+        # values seperately, due to the weighing
+        # which is not required in the evenly sampled case
+        self._steps = 0
```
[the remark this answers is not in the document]
NiMlr: I checked
`a=np.array([1.1, .2, 2.1])
 x= 0
 x+=np.sum(a)
`
in Python 2.7 and 3.5. In both it x returns as `3.4000000000000004`. But I agree setting it to be a floating point value is the more consistent and robust solution. I adapted it accordingly.

mdp/utils/covariance.py
```diff
@@ -133,6 +135,120 @@ def fix(self, center=True):
...
+        :type bool: bool
+
+        :returns: Generalised covariance matrx, average and length
+        of sequence (in time/ summed increments).
+        :rtype: Tuple[np.ndarray, np.ndarray, np.ndarray]
+        """
+        # local variables
+        type_ = self._dtype
+        steps = self._steps
+        _check_roundoff(steps, type_)
```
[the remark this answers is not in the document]
NiMlr: Done.

mdp/test/test_Unevenly_Sampled_Covariance.py
NiMlr: Subtracting reduces the precision here, as it is the "-exponent". In this case it was kind of hard for me to find a reasonable value. I set the value such that the test still returns a consistent result with "reasonable probability".

mdp/utils/covariance.py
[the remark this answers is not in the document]
Stewori: Interesting. So far I thought `+=` couldn't change the type of a variable. Proved me wrong...

mdp/test/test_UnevenlySampledSFANode.py
NiMlr: Would not it only be close to the same result? As we still approximate by using the trapezoid rule as opposed to a step function.

I will implement this. Should we also invest the time to adapt the original SFA tests to work with this algorithm?

mdp/utils/covariance.py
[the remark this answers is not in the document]
Stewori: Indeed:
```
stefan@stefan-y260 ~ $ python
Python 2.7.12 (default, Nov 12 2018, 14:36:49) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> x = 0
>>> type(x)
<type 'int'>
>>> id(x)
13939056
>>> x += .5
>>> x
0.5
>>> type(x)
<type 'float'>
>>> id(x)
14010040
>>> y = .0
>>> type(y)
<type 'float'>
>>> id(y)
14010016
>>> y += .5
>>> y
0.5
>>> type(y)
<type 'float'>
>>> id(y)
14009992
>>> 
```
Okay, ints and floats are not mutable and are always replaced by new instances on write access. That means, for a float or int, `+=` is actually not an in-place operation. If this operation happens very frequently it may be favorable to use a 1x1 numpy array instead so we get an actual inplace addition, saving lots of memory allocations.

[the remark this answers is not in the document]
NiMlr: `import numpy as np
import time

n = 1000000

t = 0.
dt = 0.1
start = time.time()
for i in range(n):
    t += dt
print(time.time()-start)

t = np.array(0.)
dt = np.array(0.1)
start = time.time()
for i in range(n):
    t += dt
print(time.time()-start)`
`0.12778997421264648
2.427353858947754`

mdp/nodes/sfa_nodes.py
NiMlr: The same thing counts for the derivative I believe. But this topic might call for a practical study on the effects of different approximation schemes in "typical" SFA problems. There could also be a generalized SFA with approximation schemes as a parameter.

mdp/utils/covariance.py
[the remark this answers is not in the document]
Stewori: Looks like Python's optimization is able to compete with numpy arrays for ordinary loops and a single element. Things should look different when using np.sum instead of the loop.
Anyway, it also occurred to me that the += on the int happens only once per chunk which is pretty seldom, so it's really the best to stick to the original implementation.
If you like, you can restore the original implementation without explicit `_tlen = 0.`.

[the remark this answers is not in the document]
NiMlr: Is there a meaningful advantage? Would not intentions be more clear this way or do we rather save a line?

mdp/test/test_Unevenly_Sampled_Covariance.py
Stewori: Oh sorry, misread it as e-3 and thought it was the tolearance rather than the precision. In that case, -5 looks like pretty low precision, we have `decimal == 7`, right? Is this due to trapezoidal rule vs ordinary SFA?

mdp/utils/covariance.py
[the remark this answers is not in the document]
Stewori: This is just following the maxim to minimize redundant code. Since int and float are better compatible than I suspected, the reinitialization is not required. I remember some issues of this kind regarding numpy arrays, that's why I was suspicious here. Apparently that was a distinct situation.

mdp/test/test_Unevenly_Sampled_Covariance.py
NiMlr: Please check out the new testing. It should be more clear now. In the current version we are comparing the L2 norm of the new covariance matrix estimators deviation to the deviation standard one (deviation from the signals known covariance). If they are not worse than the deviation of the standard one up to a carefully handpicked factor (1 % for non random stepsize, 20% for random stepsize), tests pass.

NiMlr:
> As a note for other readers:
> A main usecase of this node is to provide a way to deal with missing data in SFA. Missing data is a common issue blocking the use of SFA on real-world data. Actually there are better ways to treat missing data than the linear interpolation applied here. However it is crucial as a bottom line for testing potential future versions. The whole thing boils down to estimating covariance matrices in context of missing data. There is kind of a research field around this, e.g. using Lomp-Scargle periodograms. Having a way to estimate covariance matrices in context of missing data can benefit several other covariance-based MDP-nodes as well, e.g. PCA.

Considering this I suggest we rename the method to be called TrapezoidalSFA or TSFA and the covariance estimator accordingly. What do you think? It seems to be more clear and future proof.

Stewori:
My plan is to have the interpolation method selected by a parameter, like we do for [rank deficit method](https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/nodes/sfa_nodes.py#L120). This avoids to get a zoo of different SFA nodes (i.e. more than we have already). In that sense the current name is okay. Though, I'd appreciate if we could find a more catchy name, but so far nothing came up to me. Regarding the covariance estimator, I think it is so close to the original that more specialized names should be left to estimators that are really based on a fundamentally different idea (like periodograms).

mdp/test/test_UnevenlySampledSFANode.py
Stewori: If I understand it correctly, the result by Trapezoid rule only differ from the naive linear estimation by other weights at the boundary. Can we have kind of an ordinary-SFA-compatibility flag that adjusts this to yield exactly (up to floating point precision) the same result as original SFA in this case? IMO that would be the clearest and safest way to test it. It would also enable to transparently study the benefits of trapezoid rule vs ordinary SFA.

mdp/nodes/sfa_nodes.py
Stewori: If we get it stable in this new node we can certainly think about backporting various things to the plain SFA node.

mdp/nodes/sfa_nodes.py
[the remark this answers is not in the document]
Stewori: I would prefer to have `dt` optional in that sense. If the only way is to defaultt it to `dt= numx.ones(..)` then be it, but wouldn't it be better to allow `dt == None` all the way through, indicating that the normalization `whatever \= dt` can be skipped? This would avoid useless division through one in the fallback case (maybe numpy would optimize it away under the hood but I wouldn't count on it).

mdp/nodes/sfa_nodes.py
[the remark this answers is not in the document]
Stewori: Yes I think so. Maybe we wouldn't even need an `__init__` in `UnevenlySampledSFANode` at all, but I am not sure right now how Python's OOP would handle that. There are to behaviours I can think of:

* it wouldn't call `__init__` at all
* it would call the the first `__init__` it can find int the mro, i.e. in `__mro__`.

The second case would mean that it would behave like a plain `super(UnevenlySampledSFANode, self).__init__(input_dim, output_dim, dtype, include_last_sample, rank_deficit_method)`-call if we omit `__init__` in `UnevenlySampledSFANode`. If that's the case we should actually omit it.

mdp/nodes/sfa_nodes.py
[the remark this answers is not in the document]
NiMlr: I think, we then would have to call the covariance methods with `dt=None` and within these we would need a handling of this input which would be non trivial. We could also call the standard covariance methods - which would be the standard SFA, but I believe the most straight forward way would be setting `dt=numx.ones(..)` and inputting that into the covariance function. Especially considering that the unevenly sampled covariance called with `dt=numx.ones(..)` is not exactly the same as the standard covariance.

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +376,58 @@ def get_eta_values(self, t=1):
...
+        :type include_last_sample: bool
+        """
+        if include_last_sample is None:
+            include_last_sample = self._include_last_sample
+        # works because x[:None] == x[:]
+        last_sample_index = None if include_last_sample else -1
+
+        # update the covariance matrices
+        self._cov_mtx.update(x[:last_sample_index, :], dt)
+        self._dcov_mtx.update(self.time_derivative(x, dt), dt[:-1])
```
Stewori: Please clarify how `dt` should be interpreted w.r.t. `include_last_sample`. In line 427, shouldn't it be `dt[:last_sample_index]`?

mdp/utils/covariance.py
```diff
@@ -133,6 +135,119 @@ def fix(self, center=True):
...
+        xcpy = numx.multiply(x[1:, :], numx.sqrt(dt/2.)[:, None])
+        self._cov_mtx += mdp.utils.mult(xcpy.T, xcpy)
+        numx.multiply(x[:-1, :], numx.sqrt(dt/2.)[:, None], out=xcpy)
+        self._cov_mtx += mdp.utils.mult(xcpy.T, xcpy)
+
+        numx.multiply(x[:-1, :], (dt/2.)[:, None], out=xcpy)
+        self._avg += xcpy.sum(axis=0)
+
+        numx.multiply(x[1:, :], (dt/2.)[:, None], out=xcpy)
+        self._avg += xcpy.sum(axis=0)
```
Stewori: The lines
```
numx.multiply(x[:-1, :], (dt/2.)[:, None], out=xcpy)
self._avg += xcpy.sum(axis=0)

numx.multiply(x[1:, :], (dt/2.)[:, None], out=xcpy)
self._avg += xcpy.sum(axis=0)
```
can be optimized as follows (due to linearity):
```
numx.multiply(x[:-1, :], dt[:, None], out=xcpy)
self._avg += xcpy.sum(axis=0)/2.

numx.multiply(x[1:, :], dt[:, None], out=xcpy)
self._avg += xcpy.sum(axis=0)/2.
```
This avoids that `(dt/2.)` allocates a numpy-internal intermediate copy and also saves lots of `/2.` divisions.

Why is `[:, None]` necessary? Wouldn't it be easier if `dt` had `len(dt.shape) == 1`?

mdp/utils/covariance.py
```diff
@@ -133,6 +135,119 @@ def fix(self, center=True):
...
+        """
+        if self._cov_mtx is None:
+            self._init_internals(x)
+        # cast input
+        x = mdp.utils.refcast(x, self._dtype)
+        # update the covariance matrix, the average and the number of
+        # observations
+        # the implementation is analogous to the evenly sampled case
+        # values might get big as it is only normalized in fix
+        xcpy = numx.multiply(x[1:, :], numx.sqrt(dt/2.)[:, None])
```
Stewori: Explicitly store `numx.sqrt(dt/2.)` in some variable and reuse it below. Internally, numpy allocates a new array for this anyway.

mdp/nodes/sfa_nodes.py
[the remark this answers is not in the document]
Stewori: I looked into [`update`](https://github.com/mdp-toolkit/mdp-toolkit/pull/48/files#diff-c45ae376d2565472569ce09e2840b995R169) and think it's not too difficult to make it work with `dt==None`:
At the beginning, set
```
_dt = 1. if dt is None else dt[:, None]
```
then later always use `_dt` in place of `dt` or `dt[:, None]`. This even improves readability as it hides `[:, None]`. Using a plain `1.` should go through with numpy broadcasting and is already cheaper than creating `ones`. I think `update` is the only place that requires further adjustment, or do I overlook something?

mdp/test/test_UnevenlySampledSFANode.py
NiMlr: The main benefit I see in the trapezoid method is the better [rate of convergence](https://en.wikipedia.org/wiki/Riemann_sum). Although that advantage might not come into effect much in practice. 

The tests now compare the results of both current covariance methods up to floating point precision. So given nothing goes wrong inside of UnSFA, it should be what you are describing.

mdp/nodes/sfa_nodes.py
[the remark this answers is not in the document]
NiMlr: Thanks for the suggestion, the solutions I had in mind were dirtier. I implemented this and tests run fine (up to the randomized tests  - which not included in this PR -  that sometimes fail due to badly set precision. Next I will adapt the signature of the UnSFA.

mdp/utils/covariance.py
NiMlr: It is necessary, I believe, to have `dt` applied to all variables along the second dimension of `x`. The downside of dividing by 2 after summing is running the danger of significantly larger value occurring. I adapted it nevertheless, as it still seems to be the more sensible solution. We may want to keep that in mind though (I will include it in the docstrings or comments).

mdp/nodes/sfa_nodes.py
NiMlr: Both should be shorter, I believe. What do you think should be the default behavior in multiple training phases with the trapezoid method? I think the chunks should be treated as separate integrals and added up.

mdp/utils/covariance.py
NiMlr: Done.

mdp/nodes/sfa_nodes.py
Stewori: I assume you speak of "multiple chunks" when you say "multiple training phases", which are two things in MDP (SFA does not support or require multiple training phases in sense of multiple runs over the same data AFAIK).
I agree the integrals of multiple chunks should be added up, but ideally with an overlap of one frame (I think the dt of this particular frame shouldn't matter), i.e. such that the overall result is equal to computing one big chunk. I know this only works if the user provides the data with the right overlap. But *if* the user gets it right the node should yield the same result as one would get by using only a single chunk. This idea should IMO also form the way to test multi-chunk processing.

mdp/utils/covariance.py
Stewori: I once was also concerned with the much larger value occurring when I looked into the `CovarianceMatrix` class and saw that it adds up all values and only divides by length in the `fix` method. However it doesn't appear to be an issue there, although I am not floating point expert enough to make a thoroughly educated statement. For now I would simply trust the authors of `CovarianceMatrix` that the large value is good enough. All the many intermediate divisions would as well increase the floating point errors. To really tell the difference we would have to test a case where an analytical solution is known.

NiMlr: Check out the first entry [here](http://mdp-toolkit.sourceforge.net/additional_utilities.html#additional-utilities).

mdp/nodes/sfa_nodes.py
NiMlr: In the most recent implementation I "linked" the chunks, essentially treating them as one integral. With optional specification of the inbetween timestep. If it is not supplied I take the mean of the bounding time steps. I also implemented a test, for checking, that the method returns the same as when supplied one big chunk. This however is not in line, if I understand correctly, with the current implementation of sfa described in the docstring of the init method. Using a the current integral approximation in sfa/covariance it does not matter, but the trapezoid method treats endpoints differently, thus it makes a difference.

otizonaizit:
@NiMlr : can this be merged or are you still working on it?

NiMlr:
The code itself was ready merge from my perspective. @Stewori suggested in [Issue 55](https://github.com/mdp-toolkit/mdp-toolkit/issues/55#issuecomment-558304516) that he wants to give the docs another review before merging. Until when should it be available for the release, at the very latest?

mdp/nodes/sfa_nodes.py
```diff
@@ -267,25 +273,26 @@ def _stop_training(self, debug=False):
...
 
-        #### solve the generalized eigenvalue problem
+        # solve the generalized eigenvalue problem
         # the eigenvalues are already ordered in ascending order
         try:
             try:
                 # We first try to fulfill the extended signature described
                 # in mdp.utils.symeig_semidefinite
                 self.d, self.sf = self._symeig(
-                        self.dcov_mtx, self.cov_mtx, True, "on", rng,
```
Stewori: From https://www.python.org/dev/peps/pep-0008/#indentation: "When using a hanging indent [...] further indentation should be used to clearly distinguish itself as a continuation line."

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +376,170 @@ def get_eta_values(self, t=1):
         return self._refcast(t / (2 * numx.pi) * numx.sqrt(self.d))
 
 
+class VartimeSFANode(SFANode):
+    """
+    Extract the slowly varying components from the input data.
+    This node can be understood as a generalization to the *SFANode*.
+
+    In particular this nodes numerically computes the integrals involved in
```
Stewori: comma after "particular" and "nodes" shouldn't be plural I guess. (otherwise "computes" is wrong)

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +376,170 @@ def get_eta_values(self, t=1):
...
+    .. admonition:: Reference
+
+        More information about Slow Feature Analysis can be found in
+        Wiskott, L. and Sejnowski, T.J., Slow Feature Analysis: Unsupervised
+        Learning of Invariances, Neural Computation, 14(4):715-770 (2002).
+    """
+
+    def __init__(self, input_dim=None, output_dim=None, dtype=None,
+                 rank_deficit_method='none'):
+        """"
```
Stewori: Looks like one " too much

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +376,170 @@ def get_eta_values(self, t=1):
...
+               sfa = SFANode()
+               sfa.set_rank_deficit_method('pca')
+
+            After such an adjustment you can run ``stop_training()`` again,
+            which would save a potentially time-consuming rerun of all
+            ``train()`` calls.
+        :type rank_deficit_method: str
+        """
+        super(VartimeSFANode, self).__init__(input_dim,
+                                             output_dim, dtype, include_last_sample=True,
```
Stewori: I would prefer to have all args in a single second line with two additional indents (i.e. 8 spaces)
I suppose the blan line before super can go away then.

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +376,170 @@ def get_eta_values(self, t=1):
...
+                                             rank_deficit_method=rank_deficit_method)
+
+        self.tphase = 0
```
Stewori: It looks like you use tphases as chunks-counter. Please fix the clash between chunks and phases (also in the doc).

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +376,170 @@ def get_eta_values(self, t=1):
...
+        :rtype: numpy.ndarray
+        """
+        outlen = x.shape[0] - \
+            1 if self.tphase == 0 or dt is None or dt.shape[0] != x.shape[0] else x.shape[0]
+        out = numx.empty([outlen, x.shape[1]])
+
+        if dt is not None:
+            # Improvements can be made, by interpolating polynomials
+            out[-x.shape[0]+1:, :] = (x[1:, :]-x[:-1, :]) / \
+                dt[-x.shape[0]+1:, None]
```
Stewori: In such a case I think it's best readable if you break the line after `=`. Also, please give the continuation line an additional indent to distinguish it from usual blocks.

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +376,170 @@ def get_eta_values(self, t=1):
...
+
+        :param x: The time series data.
+        :type x: numpy.ndarray
+
+        :param dt: Sequence of time increments between vectors. Must be
+            of length *x.shape[0]-1*. In case the training is done in multiple
+            phases with intended time dependence , *dt* should be of length
+            *x.shape[0]* starting with the second call. If it continues to be
+            of length *x.shape[0]-1* time dependence is omitted. If *dt* is
+            omitted entirely there will be no time dependence between calls and
```
Stewori: Should be "between samples" as calls only separate chunks.

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +376,170 @@ def get_eta_values(self, t=1):
...
+            it will be considered to be one everywhere within a call.
```
Stewori: I find the whole explanation hard to understand. Better write it in cases-style:
`len(dt) == x.shape[0]`: `dt[0]` will be used as time increment since previous chunk (ignored on fist call)
`len(dt) == x.shape[0]-1`: No time derivative to previous chunk considered.
`dt is None`: Time increment is assumed as constantly one (like in ordinary SFA).

mdp/nodes/sfa_nodes.py
```diff
@@ -435,7 +606,7 @@ def _train(self, x, include_last_sample=None):
 
     def _set_range(self):
         if (self.output_dim is not None) and (
-            self.output_dim <= self._expnode.output_dim):
+                self.output_dim <= self._expnode.output_dim):
```
Stewori: like mentioned earlier, please restore the additional indent of continuation lines.

mdp/nodes/sfa_nodes.py
```diff
@@ -498,25 +669,24 @@ def get_quadratic_form(self, nr):
...
-##     return deriv
+# def time_derivative(self, x):
+# rows = x.shape[0]
+# columns = x.shape[1]
+# deriv = numx.zeros((rows-1, columns), dtype=self.dtype)
+
+# weave.inline(_TDERIVATIVE_1ORDER_CCODE,['rows','columns','deriv','x'],
+# type_factories = weave.blitz_tools.blitz_type_factories,
+# compiler='gcc',extra_compile_args=['-O3']);
+# return deriv
```
Stewori: Why remove the indentation in this doc? It looked okay to me.

mdp/utils/covariance.py
```diff
@@ -133,6 +135,146 @@ def fix(self, center=True):
...
+    covariance matrix is then computed as a (centered) scalar product
+    between functions, that is sampled unevenly, using the trapezoid rule.
+
+    The mean is computed using the trapezoid rule as well.
+
+    Note that the internal sum is a standard __add__ operation.
+    """
+
+    def __init__(self, dtype=None):
+        """Initialize an *VartimeCovarianceMatrix*.
```
Stewori: an -> a

mdp/nodes/sfa_nodes.py
NiMlr: So we disregard maximum line lengths? The autopep8 formatting in VS Code came up with some of the formatting by the way. In some cases it keeps me from formatting it differently, so I will have to adjust it. In some cases the strong identiation and the maximum line-length requirement in the PEP 8 specification contradict each other. Sometimes even when there is only one argument per line.

mdp/utils/covariance.py
```diff
@@ -133,6 +135,146 @@ def fix(self, center=True):
...
+            of length *x.shape[0]-1* in case of a single call. In case of
+            multiple calls with intended time dependence - as opposed to
+            computing the results as a weighted avarage of separate calls -
+            *dt* must be of length *x.shape[0]* starting with the second call.
+            The first element will be considered as the time difference
+            between the last element of *x* in the preceding call and the
+            first element of *x* in the current call. In case of not supplying
+            a *x.shape[0]*-th element in *dt*, time dependence will be dropped.
+            If *dt* omitted entirely, it will be considered to be one everywhere
+            whereas time dependence between calls is still dropped.
```
Stewori: Like mentioned earlier, better write this in cases-style. That said, there should be an explicit way to control whether or not inter-chunk derivative is computed. Can we have an additional arg to control this behavior? E.g.
`continuous == True`: Compute derivative to previous chunk (default)
`continuous == False`: Dismiss derivative to previous chunk

I think even in the first call, if `continuous == True` there would be an interpretation for `dt[0]` if `len(dt) == x.shape[0]`: for weightinh the fist sample w.r.t. to trapezoid rule.
I have a bit mixed feelings about how to deal with inconsistent values for `continuous` and `len(dt)`.
For sake of usability I think I'd prefer it lie this:
`continuous==True and len(dt) == x.shape[0]`: Okay. If in first call, ignore `dt[0]` or use it for weighting.
`continuous==True and len(dt) == x.shape[0]-1`: Raise appropriate exception.
`continuous==False and len(dt) == x.shape[0]`: Ignore `dt[0]`. Raise warning unless in first call.
`continuous==False and len(dt) == x.shape[0]-1`: Okay.
Other length of `dt`:  Raise appropriate exception.
If `dt` is a number, i.e. has no length: Use this number as constant time increment. Consider time derivative to previous chunk according to value of `continuous`.
`dt is None`: Like previous case with `dt=1.` (default)

Stewori: Please forward this behavior to the train method of VartimeSFA.

mdp/utils/covariance.py
```diff
@@ -133,6 +135,146 @@ def fix(self, center=True):
...
+        :type dt: numpy.ndarray
+        """
+        if self._cov_mtx is None:
+            self._init_internals(x)
+        # cast input
+        x = mdp.utils.refcast(x, self._dtype)
+
+        # account for the gap between training phases
```
Stewori: training phases -> chunks

mdp/utils/covariance.py
```diff
@@ -133,6 +135,146 @@ def fix(self, center=True):
...
+        if self.tphase > 0:
+            if dt is not None and dt.shape[0] == x.shape[0]:
+                self._avg += (self.xlast + x[0, :])*dt[0]/2.
+                self._cov_mtx += (numx.outer(self.xlast, self.xlast) +
+                                  numx.outer(x[0, :], x[0, :]))*dt[0]/2.
+                self._tlen += dt[0]
+
+        # keep last observation for multiple training phases
```
Stewori: multiple training phases ->  continuation to future chunk

mdp/utils/covariance.py
```diff
@@ -133,6 +135,146 @@ def fix(self, center=True):
...
+        self._avg += xcpy.sum(axis=0)/2.
+        # should the x be reset to their original value, as they are not passed as a copy?
+        self._tlen += _dt.sum() if dt is not None else x.shape[0]-1
+        self._steps += x.shape[0]
+        self.tphase += 1
+
+    def fix(self, center=True):
+        """Returns a triple containing the generalised
+        covariance matrix, the average and length of the
+        sequence (in time/ summed increments).
```
Stewori: Either "(in time / summed increments)" or "(in time/summed increments)"

mdp/utils/covariance.py
```diff
@@ -133,6 +135,146 @@ def fix(self, center=True):
...
+
+        The covariance matrix is then reset to a zero-state.
```
Stewori: a zero-state -> zero

mdp/utils/covariance.py
```diff
@@ -133,6 +135,146 @@ def fix(self, center=True):
...
+
+        :param center: If center is false, the returned matrix is the matrix
+            of the second moments, i.e. the covariance matrix of the data
+            without subtracting the mean.
+        :type bool: bool
+
+        :returns: Generalised covariance matrix, average and length
+        of sequence (in time/ summed increments).
```
Stewori: like above

mdp/nodes/sfa_nodes.py
Stewori: Personally I find the max line length of PEP 8 too restrictive, especially given that modern screens tend to be much wider than high. Sure there should be a max line length, but a bit larger than PEP 8, e.g. 100. However, changing MDP style rules beyond PEP 8 is another story and I did not mean to violate PEP 8 here. I simply wanted to go from

```
super(VartimeSFANode, self).__init__(input_dim,
                                     output_dim, dtype, include_last_sample=True,
                                     rank_deficit_method=rank_deficit_method)
```

to

```
super(VartimeSFANode, self).__init__(
        input_dim, output_dim, dtype, include_last_sample=True, rank_deficit_method=rank_deficit_method)
```

and thought it would fit. Apparently, it doesn't. Still, the first version looks strange to me. I would personally prefer

```
super(VartimeSFANode, self).__init__(input_dim, output_dim, dtype,
        include_last_sample=True, rank_deficit_method=rank_deficit_method)
```
but that violates PEP 8 (can we consider to permit this style for MDP anyway?). So let's make it

```
super(VartimeSFANode, self).__init__(
        input_dim, output_dim, dtype, include_last_sample=True, 
        rank_deficit_method=rank_deficit_method)
```
for now.

mdp/utils/covariance.py
```diff
@@ -177,33 +177,46 @@ def update(self, x, dt=None):
...
-            a *x.shape[0]*-th element in *dt*, time dependence will be dropped.
-            If *dt* omitted entirely, it will be considered to be one everywhere
-            whereas time dependence between calls is still dropped.
+        :param dt: Sequence of time increments between vectors. 
+
+            Usage with only single chunk of data:
+                *dt* must be of length *x.shape[0]-1*.
+
+            Usage with multiple chunks of data with intended time dependence:
+                *dt* should have length *x.shape[0]-1* in the first call and
```
Stewori: Please make this less strict (also in the case of just a single chunk). Allow `len(dt) == x.shape[0]` also in first call and ignore the additional element. Raise a warning then. This gives the user more flexibility in handling, e.g. allows to reuse the same array for data buffering without case distinction for the first call and additional slicing.

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +378,200 @@ def get_eta_values(self, t=1):
...
+        if dt is not None:
+            # Improvements can be made, by interpolating polynomials
+            out[-x.shape[0]+1:, :] =\
+                    (x[1:, :]-x[:-1, :]) / dt[-x.shape[0]+1:, None]
+        else:
+            # trivial fallback
+            out[-x.shape[0]+1:, :] = x[1:, :]-x[:-1, :]
+
+        if self.tchunk > 0 and dt is not None and dt.shape[0] == x.shape[0]:
+            # check if str before checking, to avoid comparison warnings
```
Stewori: This comment can be removed as we don't compare to str any more, right?

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +378,200 @@ def get_eta_values(self, t=1):
...
+            out[0, :] = numx.divide(x[0, :]-self.xlast, dt[0])
+
+        self.xlast = x[-1, :].copy()
+        self.dtlast = dt[-1].copy() if dt is not None else 1.
```
Stewori: `self.dtlast` seems to be never used ecxept in this line. Why do we store it then?

mdp/nodes/sfa_nodes.py
```diff
@@ -369,6 +378,200 @@ def get_eta_values(self, t=1):
         return self._refcast(t / (2 * numx.pi) * numx.sqrt(self.d))
 
 
+class VartimeSFANode(SFANode):
+    """
+    Extract the slowly varying components from the input data.
```
Stewori: I also saw docs like
```
"""blah
blah blah
"""
```
in mdp, i.e. opposed to
```
"""
blah
blahblah
"""
Which one do we want? We should decide for one variant and apply it consistently.

Stewori: I think the answer is here: https://www.python.org/dev/peps/pep-0008/#documentation-strings

mdp/utils/covariance.py
NiMlr: I vote against another parameter, as then - like you describe - dt and the parameter need to be matched to have a valid input. That is why originally I supplied the "special" dt's separately. That way it is an explicit argument to a certain use case and nothing needs to be matched. I still prefer the current implementation over an extra argument.

Stewori: That would mean that length of dt is the only way to control time dependence between chunks. That's a rather subtle way and may easily lead to unexpected or inaccurate behavior. Maybe the effect is small if chunks are large, but quantifying the effect, especially given unknown usecases is another story/beyond the goal here.
Also, it uglifies the user's code: Instead of setting a flag if the user wants to change this mode of operation, slicing is necessary - harder to read, less expressive. Imagine someone reads code using this feature: Without looking into the MDP doc carefully -- really in low level detail --, it would be hard to understand what chunks etc means. An explicit argument is a much clearer pointer that there is this mode of operation to consider. Slicing details of dt in user code is a weak eye-opener that there may be more semantics to it. Making this mode of operation explicit is the safe way.

Secondly: Imagine someone needs to use this node with constant time inc (e.g. because non-constant time incs only occur in a minority of the chunks). With current usage there is no way to have time dependence between chunks while using e.g. `dt=1.2` rather than dt being an array. Yes, you can set `dt=numx.ones(whatever)*1.2` but that complicates usage and is less performant - consider large chunks, this wastes memory too.

Lastly, adding another arg is not harmful because it would have a default value. I guess most usually, time dependence is the reasonable behavior so it should be the default. That's another issue with the current version: E.g. `dt=1.2` behaves against this default.

mdp/utils/covariance.py
Stewori: This was a minor cosmetical suggestion. Keep it as it is. I don't really care.

mdp/utils/covariance.py
Stewori: Okay.

mdp/nodes/sfa_nodes.py
Stewori: I hope we won't need it again for some reason.

mdp/utils/covariance.py
NiMlr: Maybe it was used to differentiate to the integer 0.

Stewori: Yes, maybe. `zero-state` also underlines the mutable nature of the object. Keep it like it was.

Issue #50, opened 2019-07-03
Some Travis runs take very long


Stewori:
In recent Travis run https://travis-ci.org/mdp-toolkit/mdp-toolkit/builds/553703911 I observed that the first two specific configurations took about 14-15 minutes to complete while all other runs took only about 2-5 minutes.
The long-taking runs were Python 3.5 + NumPy 1.8.2 + SciKits 0.18. However a later run with Python 3.5 + NumPy 1.12.1 + SciKits 0.18 was okay so it seems to be caused by NumPy 1.8.2.
It looks like the tests were hanging most time at ``test_SFANode``.
It is difficult to see whether the log contains more useful info because it is totally spammed with the deprecation warning mentioned in #49. So we better fix that one first.
Also, given that the issuous configuration is somewhat odd this doesn't have high priority. I'd like to track it though as it might point to an issue in ``test_SFANode``.

Stewori:
Fixed now as of recent commit #53.

PR #51, opened 2019-08-02
pytest4: use tmpdir fixture

keszybz:


Stewori:
Hey @keszybz, would you drop us a note what this is fixing and how?

Stewori:
Okay, I found http://doc.pytest.org/en/latest/tmpdir.html#the-tmpdir-fixture
This is just some code cleanup without fixing a specific problem, right?
I am fine with merging this, just wanted to understand it.

Stewori:
Looked at the Travis results and several turned red. Some of them *might* be separate issues, e.g. related to #50, i.e. those failing due to hang in `test_SFANode`.
Given that there is no compelling reason for this change (is there?) I would postpone it until we understand the failures better.

Debilski:
The pytest `tmpdir` object uses the `__fspath__` protocol for compatibility with the `os.path.join` function (and others), but this protocol has only been introduced in Python 3.6, so it breaks on 3.5 unless you also rewrite the join functions. ( https://docs.python.org/3/whatsnew/3.6.html#pep-519-adding-a-file-system-path-protocol )

Stewori:
Okay, thanks for the clarification. So, I do not see a sufficiently compelling reason for the change that justifies breaking Python 3.5 compatibility. Maybe a fallback would be feasible, but I neither see a sufficiently compelling reason to invest the work to get this change in. @keszybz If you see a reasonably efficient solution, feel free to adjust the PR such that travis turns green. Then I will happily accept it.

keszybz:
I was updating the Fedora package for mdp for new compatibility with newest pytest and python3.8. I got some error related to this code. It seemed simplest to get rid of the custom code and use the tmpdir fixture. Nevertheless, now that I rested this, the error was most likely caused by something else, and current code in master works correctly with python3.8b3 and pytest-4.6.4. So this patch is not necessary. I think it'd make sense to merge after python3.5 compatiblity is not required anymore.

Issue #49, opened 2019-07-03
DeprecationWarning: inspect.getargspec() is deprecated


Stewori:
Recent Travis logs are full of the following warning (though passing):
```
/home/travis/build/mdp-toolkit/mdp-toolkit/mdp/signal_node.py:165:
DeprecationWarning: inspect.getargspec() is deprecated,
use inspect.signature() or inspect.getfullargspec()
```
This doesn't look too difficult to fix, so we should get it done to make the logs cleaner and thus more useful.

Stewori:
Fixed in https://github.com/mdp-toolkit/mdp-toolkit/commit/9d2d1be343691eb51e84261ab52130223e748a1f

PR #52, opened 2019-08-09
Fix for "DeprecationWarning: inspect.getargspec() is deprecated"

NiMlr:
The fix adressing [issue 49](https://github.com/mdp-toolkit/mdp-toolkit/issues/49) was pretty straightforward and based on the suggested replacement "inspect.getfullargspec". The only complication I came across was python 2 requiring to use the old method. So I added a handling for the module level imports.

### DIFF
--- a/bimdp/binode.py
+++ b/bimdp/binode.py
@@ -73,7 +73,11 @@ keywords are treated in a special way:
 from builtins import next
 from builtins import str
 
-import inspect
+# python 2/3 compatibility
+try:
+    from inspect import getfullargspec as getargs
+except ImportError:
+    from inspect import getargspec as getargs
 
 import mdp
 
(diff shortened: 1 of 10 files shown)


Stewori:
Hmm, some travis runs are red. This looks like a manifestation of #50. I am somewhat sure that this is not caused by this fix.

NiMlr:
I also believe these changes should not affect the runtime of the tests strongly. Restarting the jobs through the Travis GUI (requires maintainer access) could make them pass, as tests with the same version combo [passed in my repo](https://travis-ci.org/mdp-toolkit/mdp-toolkit/builds/569620020?utm_source=github_status&utm_medium=notification).

I can tackle this issue separately next if that helps.

Stewori:
Okay, thanks for fixing #50. Then I suppose this can go in.

PR #53, opened 2019-08-10
FIX for issue 50: Update .travis.yml

NiMlr:
It is not explicitly stated, which Python versions are supported by Numpy 1.8.2, but Numpy 1.8.0 supports only 2.6 -2.7/3.2 - 3.3 and Numpy 1.9.0 supports only 2.6 - 2.7 and 3.2 - 3.4 [1]. So I propose to exclude testing on this version combination to address [issue 50](https://github.com/mdp-toolkit/mdp-toolkit/issues/50).

[1] https://docs.scipy.org/doc/numpy/release.html


Stewori:
Thanks, that sounds like a reasonable solution to me.

PR #54, opened 2019-08-12
FIX: Precision on test_ccipcanode_v2.

NiMlr:
The precision on this test seems to be set to sharp, as it keeps failing.

### DIFF
--- a/mdp/test/test_CCIPCANode.py
+++ b/mdp/test/test_CCIPCANode.py
@@ -67,7 +67,7 @@ def test_ccipcanode_v2():
 
     print('\nTotal Time for {} iterations: {}'.format(
         iterval, time.time() - _tcnt))
-    assert_almost_equal(numx.ones(output_dim), dcosines[-1], decimal=2)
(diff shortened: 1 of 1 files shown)


NiMlr:
I did a test using the repo [at this point in time](https://github.com/mdp-toolkit/mdp-toolkit/tree/5ffe1df89156ff84a2927da7c4304f1512f95c88). Namely, I ran `test_ccipcanode_v2` 400 times. It failed 9 times. As we currently have 26 jobs running in Travis and given they are independent, to reach a 0.8 probability of a success of a travis run, we would need to have a probability of a single test succeeding  of at least `q = (0.8)^(1/26) =  0.9914543`. Assuming we have a probability of the test succeeding of at least that, the probability of observing 9 tests or more out of 400 failing is smaller than
<p align="center">

</p>

Stewori:
This time test_inverse failed. Still I will check this in now in the hope that no random test fails again. Keep fingers crossed...
Nice math btw. We should pin your thoughts as a separate issue.

PR #65, opened 2019-12-14
Pytest 5

Stewori:
This PR is not intended for merging but only to get an online-visible reference of breakages with pytest 5.


otizonaizit:
Fixed in #74

Issue #61, opened 2019-12-13
Warnings in tests


NiMlr:
I collected a list of warnings popping up in different tests. Some may be relevant only to some version combinations of the dependencies.

#### MDP
- [x] `DeprecationWarning: object of type <class 'float'> cannot be safely interpreted as an integer.`
appearing in [test_EtaComputerNode.py line 6](https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/test/test_EtaComputerNode.py#L6) in this [test run](https://travis-ci.org/mdp-toolkit/mdp-toolkit/jobs/624784648?utm_medium=notification&utm_source=github_status).

- [x] ``DeprecationWarning: `formatargspec` is deprecated since Python 3.5. Use `signature` and the `Signature` object directly``
appearing in [signal_node.py line 180](https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/signal_node.py#L180) in this [test run](https://travis-ci.org/mdp-toolkit/mdp-toolkit/jobs/624784645?utm_medium=notification&utm_source=github_status). There are more instances in the code where `formatargspec` is used.

- [x] `FutureWarning: Default solver will be changed to 'lbfgs' in 0.22. Specify a solver to silence this warning.`
appearing in tests of scikit-learn-nodes in [test_nodes_generic.py](https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/test/test_nodes_generic.py#L193) in this [test run](https://travis-ci.org/mdp-toolkit/mdp-toolkit/jobs/624784645?utm_medium=notification&utm_source=github_status).

- [x] ` FutureWarning: The default value of cv will change from 3 to 5 in version 0.22. Specify it explicitly to silence this warning.`
appearing in tests of scikit-learn-nodes in [test_nodes_generic.py](https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/test/test_nodes_generic.py#L193) in this [test run](https://travis-ci.org/mdp-toolkit/mdp-toolkit/jobs/624784645?utm_medium=notification&utm_source=github_status).

- [x] `ConvergenceWarning: Liblinear failed to converge, increase the number of iterations.`
[...]

PR #67, opened 2019-12-15
A first effort in tackling the warnings described in issue 64

NiMlr:
See first and second entry in [issue 64](https://github.com/mdp-toolkit/mdp-toolkit/issues/61).


NiMlr:
The second fix is copied from [this code](https://github.com/nltk/nltk/blob/81490305bbbb42c6b7fa62a4694d65607b7f420f/nltk/decorators.py#L28). I am not sure which part of MDP is best to include the legacy adaption code, as it is needed for all occurrences of `formatargspec`. Note, that I have not fixed all of these occurrences yet.

mdp/utils/routines.py
```diff
@@ -11,6 +11,22 @@
...
+    Basically, from a Signature we create a string and remove the default values.
+    """
+    listsignature = str(signature)[1:-1].split(",")
+    for counter, param in enumerate(listsignature):
+        if param.count("=") > 0:
+            listsignature[counter] = param[0:param.index("=")].strip()
+        else:
+            listsignature[counter] = param.strip()
+    return ", ".join(listsignature)
+
```
Stewori: I think on top-level PEP 8 suggests two lines to separate functions. I've seen this pattern frequently in MDP so far.

mdp/utils/routines.py
```diff
@@ -11,6 +11,22 @@
 import random
 import itertools
 
+def _legacysignature(signature):
+    """
+    Used to replace inspect.formatargspec by inspect.signature.
```
Stewori: ->    `"""Used to replace inspect.formatargspec `

mdp/signal_node.py
```diff
@@ -187,7 +174,14 @@ def _function_infodict(func):
             argnames.append(varargs)
         if varkwargs:
             argnames.append(varkwargs)
-        signature = _legacysignature(inspect.signature(func))
+        try:
+            signature = _legacysignature(inspect.signature(func))
+        except:
```
Stewori: Do we know the raised exception more specifically? I think a plain `except` should be avoided if reasonably possible.

Stewori:
Difficult to tell where it should go.
I see these possibilities: configuration.py, helper_funcs.py, utils/introspection.py or make a custom file, e.g. utils/compatibility.py
Maybe helper_funcs would be best?

Stewori:
And thanks for this work. I'd be happy to include it into the upcoming release.

NiMlr:
> Difficult to tell where it should go.
> I see these possibilities: configuration.py, helper_funcs.py, utils/introspection.py or make a custom file, e.g. utils/compatibility.py
> Maybe helper_funcs would be best?

Based from the contents of the files, I think it fits better in routine.py than in helper_funcs.py.

mdp/signal_node.py
NiMlr: Yes, you are right. It should be an `AttributeError`. See [build](https://travis-ci.org/mdp-toolkit/mdp-toolkit/jobs/625163939#L4622).

Stewori:
It appeared to me that routines mostly consists of data-processing routines. However, the same is true for helper_funcs but the name sounded more generic to me. Anyway, I don't really care. routines is fine. Please separate all top-level functions in routines by two blank lines if you put it there. Thanks.

NiMlr:
> It appeared to me that routines mostly consists of data-processing routines. However, the same is true for helper_funcs but the name sounded more generic to me. Anyway, I don't really care. routines is fine. Please separate all top-level functions in routines by two blank lines if you put it there. Thanks.

Should be included in the most recent commit.

mdp/signal_node.py
```diff
@@ -173,10 +174,15 @@ def _function_infodict(func):
...
-                                          defaults,
+
+        # python 2/3 compatibility
+        try:
+            signature = _legacysignature(inspect.signature(func))
+        except AttributeError:
+            signature = inspect.formatargspec(regargs,	      
+                                          varargs,	
+                                          varkwargs,	
+                                          defaults,
```
Stewori: Either fix the alignment or put it like this:
```
signature = inspect.formatargspec(
        regargs, varargs, varkwargs, defaults, formatvalue=lambda value: "")[1:-1]
```
(additional linebreak as needed)

NiMlr: The alignment should not be different, as then the column limit is exceeded by the last argument.

Stewori: Please make it a three-liner then. This non-alignment just looks wrong.
```
signature = inspect.formatargspec(
        regargs, varargs, varkwargs, defaults,
        formatvalue=lambda value: "")[1:-1]
```

Stewori:
I will merge a traivs.yml update first because that will save 7-10min for future travis runs.

NiMlr:
I still have to fix the other occurrences of `fortmatargspec` anyway.

Stewori:
Please also add a line to CHANGES on that occasion:
`2019-15-12: FIX: Warnings type interpretation and formatargspec deprecation`

NiMlr:
There is a problem. The implementation of `_legacysignature` is not general enough in transforming the output of `inspect.signature` to the one of `inspect.formatargspec`. In order to only use the new `inspect.signature` we would need to rewrite whatever functions use it downstream - so I tried to avoid it. Others have avoided it by [implementing the functionality of `inspect.formatargspec`](https://github.com/NikolayPogoreliy/sqla_tests/commit/536d5187a038a44aec624dd2a99792f49dec82ed). I am trying to do this now.

mdp/signal_node.py
```diff
@@ -175,15 +176,10 @@ def _function_infodict(func):
...
-        try:
-            signature = _legacysignature(inspect.signature(func))
-        except AttributeError:
-            signature = inspect.formatargspec(regargs,	      
-                                          varargs,	
-                                          varkwargs,	
-                                          defaults,	
-                                          formatvalue=lambda value: "")[1:-1]
+        signature = inspect_formatargspec(
+                regargs,	varargs, varkwargs,	defaults,
```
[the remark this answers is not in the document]
NiMlr: Fixed it.

NiMlr:
The PR can merged if you are ok with it.

Stewori:
Thanks again for the good solution.

PR #73, opened 2020-03-10
ERF: adapt directive to work with epydoc

NiMlr:
I am proposing to adapt the directive ".. attribute:: atr" of the rst-docstring style to ":ivar atr:". As @Stewori pointed out, this directive works with Epydoc and lets us build the [legacy documentation](http://mdp-toolkit.sourceforge.net/api/index.html).

[Reference](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists)


keszybz:
Looks reasonable.

Stewori:
Thanks for this work!

PR #74, opened 2020-04-17
[WIP] Prepare release 3.6

otizonaizit:
This PR is to track the work done in preparation of release 3.6.

For now, I fixed incompatibilities with modern pytest versions, re-enabled tests that were skipped in the past but do not need to be skipped anymore, ignored warnings from sklearn during the testing: we are not responsible for these warnings anyway.

I have disabled a test for the GSFANode, because I have a hard time understanding what that test is doing, and the tests fails quote often. @Stewori , @NiMlr , @AlbertoEsc : do you know more about it? Can this be made more stable somehow? It does not seem to be a problem with precision.

I will add commits here in preparation of the release.


otizonaizit:
@Stewori @NiMlr : don't you think that we could reduce the test matrix a bit? it takes forever for travis to test all the stuff, and I think we could remove a lot of ancient stuff from there, no? I'd go for leaving there only:

   - Python versions: the latest Python, plus the Python version available in Debian stable (this is the most ancient thing people are going to encounter anyway)
  - I'd use the same criteria for numpy, scipy and sklearn: only the version in Debian stable plus the latest

There's the question Python 2, of course. given that numpy, scipy and sklearn stopped supporting it, I'd say that we could drop Python 2 too in this release.

otizonaizit:
In f1f114e I am implementing my proposal in comment https://github.com/mdp-toolkit/mdp-toolkit/pull/74#issuecomment-615283329

Stewori:
Re the test matrix: We widened it for the release, planning to cut it down right after the release.
Re GPFA: @AlbertoEsc is the only expert in that

otizonaizit:
Well, OK, but again, I see really absolute no point in testing all that mess of combinations of Pythons and dependencies. Most combinations don't even make sense, and the matrix had a lot of excludes exactly because of that.
What do you think of my proposed setup? In short: we test the current Debian stable (that is ancient enough to cover for current Ubuntus too), and the latest of everything, to catch errors with new features/deprecations/removals from our dependencies.

Stewori:
I think we worked out the bottom line of the test matrix (the odd cases) to be able to tell MDP's minimum requirements (e.g. for release notes). Also, the old tests were inherited and I was careful in dropping old versions because I was not aware of the state of the libs in Debian.

Re removing numx: Yes, but not for the upcoming release. The change would just be too big.

The proposed changes look good to me.

mdp/test/test_VartimeSFANode.py
```diff
@@ -69,7 +69,7 @@ def test_VartimeSFANode2():
             or ((unsfa2.sf[0, 0] < 0) and (unsfa.sf[0, 0] > 0)):
         unsfa2.sf *= -1.
 
-    assert_array_almost_equal(unsfa.sf, unsfa2.sf, decimal=10)
+    assert_array_almost_equal(abs(unsfa.sf), abs(unsfa2.sf), decimal=10)
```
NiMlr, line 72: Did the VartimeSFANode fail due to a sign? Or is it just in case? The above conditional was supposed to catch that, but it may have missed the high order features. I am asking be cause we could get rid of the above conditional in this case.

otizonaizit: The tests sometimes failes due to sign changes, yes. OK, so I'll get rid of the conditionals.

otizonaizit:
With the last series of commits I have added back a couple of tests for old versions:

  - we stay compatible for Python 2.7 and Python >= 3.5 with this release

Except for the Debian stable case, where we use the same versions as the one available in the corresponding packaging system, for everything else we assume people will use pip to upgrade to the latest available versions of all dependencies.

NiMlr:
Just for understanding: Why we ignoring the sklearn warnings? Is it due to API changes announced in older versions that turned out to not affect us?

NiMlr:
Are these not needed?
```
 parallel python: NOT AVAILABLE: No module named 'pp'
          shogun: NOT AVAILABLE: No module named 'shogun'
          libsvm: NOT AVAILABLE: No module named 'svm'
```

otizonaizit:
Because we call sklearn routines with default parameters (i.e. not hard-coded).

Those warnings are meant to warn users of sklearn of coming changes or deprecated parameters.

So, we can not do anything to fix them, nor I think it makes any sense to try and fix them.

We create a transparent layer to use sklearn algorithms in MDP as nodes. The more transparent the better, including warnings. Just, it makes no sense to litter our tests outputs with those warnings. The warnings will be shown if a MDP user is actually *using* the sklearn nodes: there is the place where this warnings are needed.
Is it clear what I mean?

otizonaizit:
No, these are ancient libraries, either not ported to Python 3 (pp) or superseded by more modern libraries (shogun and libsvm are superseded by sklearn).
After release we should drop the corresponding nodes in MDP.

NiMlr:
> Is it clear what I mean?
Clear!

> After release we should drop the corresponding nodes in MDP.
If you have the capacity and do not plan on implementing this change yourself, it would surely be helpful if we had an issue that is as detailed as possible on this.

otizonaizit:
> > After release we should drop the corresponding nodes in MDP.
> 
> If you have the capacity and do not plan on implementing this change yourself, it would surely be helpful if we had an issue that is as detailed as possible on this.

Sure, as soon as I have a clear picture I'll open issues to track all modernizations I see fit. After you asked I went back and discovered that libsvm has released indeed a version compatible with Python 3, so I have adjusted the code a bit to work with it and enabled it again! (They dropped Python 2 compatibility along the way, so that is gone).
 
> Do you have an idea of why the test in the newest versions takes roughly 4 times as long as the rest of the tests? Will it present a problem? It is not new with this PR by the way.

Do you mean the scikit-learn tests? Most probably because there are a lot of automatically generated nodes, and we test output/input dim and dtype consistency for each one of those…

NiMlr:
Great, thanks!

> Do you mean the scikit-learn tests? 

I mean the Python 3.8 test [here](https://travis-ci.org/github/mdp-toolkit/mdp-toolkit/builds/676617882).

AlbertoEsc:
Thanks a lot for letting me know that a GSFANode's test is failing frequently, I will have a look at this so that it can be re-enabled soon.
It looks like MDP is again quite active, congrats!
Enjoy the rest of the weekend

otizonaizit:
> I mean the Python 3.8 test [here](https://travis-ci.org/github/mdp-toolkit/mdp-toolkit/builds/676617882).

I really don't know. It seems the test that is taking the long (almost 7 minutes) is this one:
 https://github.com/mdp-toolkit/mdp-toolkit/blob/8105cd121030916289183595617280e01715bcf2/mdp/test/test_SFANode.py#L81

This test on my machine with Python3.8 takes 9.14 seconds. On my machine, on  Python 3.7 it takes 8.86 seconds. My diagnose is that the problem is with the travis CI setup with Pytohn 3.8. Given that the test is using LAPACK pretty heavily, I would guess there must be a problem with the openblas linked to scipy, which probably does not work properly on the travis platform for Python 3.8.

Anyway, nothing we can fix, IMHO.

otizonaizit:
So, pre-release work is done. I am merging this pull request and will tag the 3.6 release on master in a minute!

Issue #76, opened 2020-04-19
fix random seed for releases


Stewori:
I am following up on private email communication where we discussed the issue of small but notable probability of failing random tests due to insane random data.
We concluded it is best to fix a seed for releases, so the tests of a release run reliably.

Alternatives would have been to lower testing precision, but that would spoof test reliability and would not eliminate the risk of failure, just make it less likely, which is not what we want. The algorithms cannot be improved to handle every ill kind of data because in the end this relies on numerical operations and for every numerical operation there is some nasty data that provokes instability. Checking or sanetizing data would in general be unnecessarily expensive, given the low probability of failure. For frequent and annoying cases this may be an option though (c.f. [symeig_semidefinite](https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/utils/symeig_semidefinite.py)). 

A remaining question is how to proceed after the release. I propose we should unfix the seed again, but track results to learn more about the situation. I'd suggest we collect a list of sane seeds (around 1000 or so) and then randomly choose from these. Still fix a sane seed for every release.
For the upcoming release, let's just fix some good seed and unfix it afterwards.

Suggestions, opinions?

NiMlr:
Based on my observations, we have a small subset of mostly the same nodes' tests failing. This suggests we do have higher instability in single algorithms. I might be wrong but if not I would conclude that ideally, the test of a node should come with a fixed - as opposed to random - set of test data. At least unless if it is not absolutely necessary to have it randomized. I know this solution will be hard to come up with so a simpler version would be to identify the critical nodes by sampling tests and then fix them. I would be happy to support this effort.

Stewori:
I think it is valuable to collect some data to identify nodes that are specifically fragile. Actually, such nodes should be revisited to ensure it's not a subtle bug or design fail allowing for a stability improvement. I doubt we have capacity for that, but collecting the data would be a good step in any case.
Still, I would go for fixed seeds for releases. I am convinced that *every* node that is based on numerical computation (so, in fact *every* node) has some nonzero failure probability on random data. E.g., as soon as very large and very small values occur in the same data, numerical results can be arbitrarily inaccurate. This affects matrix decompositions (also regarding eigenvalues), matrix inversion, but also basics like addition (multiplication is better). This is not the only form of numerical instability.

Stewori:
[...]

PR #80, opened 2020-04-20
ERF: Loosen pytest-seed and add a file for collection broken seeds

NiMlr:
The PR is an implementation of [this suggestion](https://github.com/mdp-toolkit/mdp-toolkit/issues/76#issuecomment-616452250).

### DIFF
--- a/BROKEN_SEEDS
+++ b/BROKEN_SEEDS
@@ -1,0 +1,2 @@
+# This file is a collection of pytest-seeds that led to at least one of the
+# tests failing, and it provides a way to spot faulty functionality.
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ install:
     - pip install $DEPS
 
 script:
-    - pytest --cov-report= --cov-config=.coveragerc --cov=mdp --seed=725021957 mdp
-    - pytest --cov-report= --cov-config=.coveragerc --cov=bimdp --cov-append --seed=725021957 bimdp
+    - pytest --cov-report= --cov-config=.coveragerc --cov=mdp mdp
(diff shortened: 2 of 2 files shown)


BROKEN_SEEDS
```diff
@@ -0,0 +1,2 @@
+# This file is a collection of pytest-seeds that led to at least one of the
+# tests failing, and it provides a way to spot faulty functionality.
```
otizonaizit, line 2: I would add a clear specification of how to store the seeds. So, we should first find a broken seed so the we can add an example here. Anyway, we should explain that a failing test can be reproduced like this:
```
pytest --seed==N mdp/test/test_XXX.py -k test_YYY
```
The values for `N`, `XXX` and `YYY` can be read out of the pytest report after a failed test run.

Here an example of failing seed (from: 664d68883d2424e0ef3d6ff003ba806594f54556)

pytest --seed=2133340156 mdp/test/test_GSFANode.py -k test_basic_GSFA_edge_dict

(note that to get a failure for this specific test you need to comment out the `@pytest.mark.skip` decorator in the file)
So maybe it is enough if we just add the above line to this file. When we discover other failing seed/tests, we can add the lines to reproduce here.

NiMlr: Do you think extending the section on testing in the docs is a better place for this or should at least be considered additionally?

otizonaizit:
Yes, good idea! But in the developers guide, not in the install/testing section. That is not something users should care about.

NiMlr:
See the [development guide entry](https://mdp-toolkit.github.io/development.html#how-to-handle-failed-tests) that describes how to handle failed tests.

PR #85, opened 2020-05-15
DOC: Improved documentation in symeig_semidefinite.py.

Stewori:


mdp/nodes/sfa_nodes.py
```diff
@@ -47,12 +47,15 @@ class SFANode(Node):
     """Extract the slowly varying components from the input data.
 
     :ivar avg: Mean of the input data (available after training)
+    :vartype avg: numpy.ndarray
```
Stewori, line 50: These vartypes are just an experiment I forgot to clean up for this commit. However, it turned out well and works properly with epydoc. So I would leave them here. Also, the hassle to put them into a separate PR is not worth it.

PR #86, opened 2020-05-15
DOC: Point to the new website at https://mdp-toolkit.github.io.

Stewori:


NiMlr:
Do not forget the [__homepage__-variable](https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/__init__.py#L122)

PR #90, opened 2021-06-08
ERF: add CuPy implementation

nkgevorgyan:


mdp/configuration.py
```diff
@@ -264,20 +265,49 @@ def get_numx():
...
+    except:
+        gpu_available = False
+
     USR_LABEL = os.getenv('MDPNUMX')
 
     # check if the variable is properly set
-    if USR_LABEL and USR_LABEL not in ('numpy', 'scipy'):
+    if USR_LABEL and USR_LABEL not in ('numpy', 'scipy', 'cupy'):
         err = ("Numerical backend '%s'" % USR_LABEL +
                "not supported. Supported backends: numpy, scipy.")
```
Stewori, line 281: Here should cupy be listed too.

mdp/configuration.py
```diff
@@ -264,20 +265,49 @@ def get_numx():
...
+                raise ImportError(exc)
+            else:
+                numx_exception['cupy'] = exc
+    
+
     # if variable is not set or the user wants scipy
-    if USR_LABEL is None or USR_LABEL == 'scipy':
+    if (USR_LABEL is None or USR_LABEL == 'scipy') and numx_description is None:
         try:
+            print('scipyyyy')
```
Stewori, line 310: (cleanup...)

Stewori:
Hey @nkgevorgyan, first of all, thanks for this work! I did not yet test it with actual cuda-capable hardware, so for now just some first remarks:

- there is a test failing in SFANode. Is it due to this PR? Or did something change in a dependency (e.g. NumPy). It appears the test is failing without having cupy in operation. Clearly, existing tests should be unaffected by this PR, especially if numx!=cupy anyway.
- throughout the change set it looks like you had to build in fallbacks to numpy at several places. I guess that is for stuff that is not provided by cupy. Is that correct? I suppose the fallbacks should have a better abstraction, but I'll have to give it more thought and testing.

mdp/utils/__init__.py
```diff
@@ -44,9 +44,13 @@
 # matrix multiplication function
 # we use an alias to be able to use the wrapper for the 'gemm' Lapack
 # function in the future
-mult = _mdp.numx.dot
-matmult = mult
-
+try:
+    mult = _mdp.numx.dot
+    matmult = mult
```
Stewori, line 49: Does cupy really not have ``dot``?

mdp/utils/_symeig.py
```diff
@@ -139,8 +140,29 @@ def _symeig_fake(A, B = None, eigenvectors = True, turbo = "on", range = None,
             # diagonalize A
             w, ZA = numx_linalg.eigh(A)
             Z = mdp.utils.mult(ZB, ZA)
-    except numx_linalg.LinAlgError as exception:
-        raise SymeigException(str(exception))
+    except:
```
Stewori, line 143: What exceptions are to be expected here? I guess this is something like cupy is not supported. In that case mdp souldn't silently fallback to using numpy, but tell the user the issue (an maybe recommend using a different numx backend).

cclauss: https://realpython.com/the-most-diabolical-python-antipattern

Issue #92, opened 2021-08-19
`numpy.typeDict` is deprecated


otizonaizit:
In https://github.com/mdp-toolkit/mdp-toolkit/blob/300d288d75ac8495f7f4a0ef40cb71d4ff58a311/mdp/utils/routines.py#L295
we use `np.typeDict`. This is deprecated since forever: https://github.com/numpy/numpy/commit/668950285c407593a368336ff2e737c5da84af7d and https://github.com/numpy/numpy/issues/17585 . Since numpy release [1.21](https://github.com/numpy/numpy/releases/tag/v1.21.0) this now issues a deprecation warning: https://github.com/numpy/numpy/pull/17586
It seems that `scipy` is not importing deprecated names from the numpy namespace anymore, so that with numpy 1.21 and scipy 1.7 you now get an error:

```python
AttributeError: module 'scipy' has no attribute 'typeDict'
```

We should change `np.typeDict` with `np.sctypeDict` most probably, but it is probably better to try to understand what we were trying to do in the function `get_dtypes` and fix it properly:
https://github.com/mdp-toolkit/mdp-toolkit/blob/300d288d75ac8495f7f4a0ef40cb71d4ff58a311/mdp/utils/routines.py#L274

Thanks to Elgin Road for reporting this!

Stewori:
Thank you, Nils.

finnoshea:
Using conda and a clean install of mdp with `conda install mdp` I can `import mdp` without problem.  But, if I then install scipy (and do nothing else), when I `import mdp` I get the AttributeError mentioned in the op of this issue.  Is there a work around for using both mdp and scipy?

NiMlr:
@finnoshea Do you experience the same problem when you install the mdp version from the master branch:
```bash
pip install git+https://github.com/mdp-toolkit/mdp-toolkit
```

finnoshea:
That does appear to work.  To recap:

0) new conda env
1) conda install mdp -> import mdp works
2) conda install scipy -> import mdp fails

versus

0) new conda env
1) pip install git+https://github.com/mdp-toolkit/mdp-toolkit -> import mdp works
2) conda install scipy -> import mdp still works

Thanks for the pointer @NiMlr!

Stewori:
This issue has been fixed in #94, so I am closing it now.

PR #94, opened 2022-03-27
FIX: Deprecation of numpy.typeDict

NiMlr:
This PR addresses #92.

The proposed alternative `numpy.sctypeDict` seems to be discouraged in https://github.com/numpy/numpy/issues/17585.

Based on our relatively shallow current usage of `numpy.typeDict`:
https://github.com/mdp-toolkit/mdp-toolkit/blob/300d288d75ac8495f7f4a0ef40cb71d4ff58a311/mdp/utils/routines.py#L295

I propose to replace `numpy.typeDict[d]` by the generic `numpy.dtype(d).type` as:
```python
# excluded the invalid inputs; see below
>>> [numpy.typeDict[d]==numpy.dtype(d).type for d in ['float16', 'float128', 'complex256']]
# [True, True, True]
```

This change to a function (as opposed to dictionary) has the downside of having to exclude the check:
https://github.com/mdp-toolkit/mdp-toolkit/blob/300d288d75ac8495f7f4a0ef40cb71d4ff58a311/mdp/utils/routines.py#L297

However, we learn that exactly 2 of the current 5 checks will return `false` in the most recent release of numpy as of 03/27/22:
```python
>>> 'float96' in numpy.typeDict or 'complex192' in numpy.typeDict
# False
>>> 'float16' in numpy.typeDict and 'float128' in numpy.typeDict and 'complex256' in numpy.typeDict
# True
>>> numpy.__version__
# '1.22.3'
```
I excluded the invalid ones and dropped the check. I think it is likely this will not create problems down the line.

**General note:**
I am not sure the functionality of checking for unsafe types generically is still up-to-date. Especially as it is only run if one is not using `scipy.linalg.eigh`:
https://github.com/mdp-toolkit/mdp-toolkit/blob/300d288d75ac8495f7f4a0ef40cb71d4ff58a311/mdp/utils/routines.py#L286

### DIFF
--- a/mdp/utils/routines.py
+++ b/mdp/utils/routines.py
@@ -292,9 +292,8 @@ def get_dtypes(typecodes_key, _safe=True):
     return types
 
 
-_UNSAFE_DTYPES = [numx.typeDict[d] for d in
-                  ['float16', 'float96', 'float128', 'complex192', 'complex256']
-                  if d in numx.typeDict]
+_UNSAFE_DTYPES = [numx.dtype(d).type for d in
+                  ['float16', 'float128', 'complex256']]
 
 
 def nongeneral_svd(A, range=None, **kwargs):


Stewori:
Since this passes local tests and given that #95 requires it to pass, I will merge this now.

Issue #99, opened 2023-06-23
incompatible with numpy 1.25.0


stanislavlevin:
Current mdp's master against numpy 1.25.0:
```
[builder@localhost mdp-toolkit]$ python3 -c 'import mdp'
Traceback (most recent call last):
  File "/usr/lib64/python3.11/inspect.py", line 1369, in getfullargspec
    sig = _signature_from_callable(func,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/inspect.py", line 2521, in _signature_from_callable
    return _signature_from_builtin(sigcls, obj,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/inspect.py", line 2328, in _signature_from_builtin
    raise ValueError("no signature found for builtin {!r}".format(func))
ValueError: no signature found for builtin <function eigh at 0x7f85eea65fb0>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/src/RPM/BUILD/mdp-toolkit/mdp/__init__.py", line 133, in <module>
    utils.symeig = configuration.get_symeig(numx_linalg)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/RPM/BUILD/mdp-toolkit/mdp/configuration.py", line 338, in get_symeig
    args = getargs(numx_linalg.eigh)[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/inspect.py", line 1379, in getfullargspec
    raise TypeError('unsupported callable') from ex
TypeError: unsupported callable
```

PR #100, opened 2023-07-25
FIX: Deprecation of inspect.getfullargspec (#99)

kotopesutility:
inspect.getfullargspec is deprecated (see https://docs.python.org/3/library/inspect.html#inspect.getfullargspec) and does not support numpy_linalg.eigh for numpy >= 1.25  (see #99)

### DIFF
--- a/mdp/configuration.py
+++ b/mdp/configuration.py
@@ -4,7 +4,10 @@ from .repo_revision import get_git_revision
 import mdp
 # python 2/3 compatibility
 try:
-    from inspect import getfullargspec as getargs
+    # Migrate to signature as recommended in
+    # https://docs.python.org/3/library/inspect.html#inspect.getfullargspec
+    from inspect import signature
+    getargs = lambda func: (signature(func)._parameters, None)
 except ImportError:
     from inspect import getargspec as getargs
 import tempfile


kotopesutility:
> Thank you for this PR -- the deprecation should indeed be fixed. 

Thank you for your  reply!

>However, I doubt that it is sufficient to fix #99 but did not check. Did you test it with recent NumPy?

Yes

>I suspect, inspect will still raise `TypeError: unsupported callable`.

Well, I checked once again and my solution works. Also all tests passed. 

>I guess, https://github.com/mdp-toolkit/mdp-toolkit/blob/master/mdp/configuration.py#L338 needs to be fixed, e.g.\ by wrapping in try/except TypeError.

If you get TypeError, you still need to check if it is numpy. 

I don't like that I use a tuple with None at the end, but what I did was to patch just one line.

Stewori:
> If you get TypeError, you still need to check if it is numpy.

Yes, that's not ideal, your fix is better if it is sufficient as you say.

Stewori:
Thanks again for this PR. Given that tests pass I am going to merge it soon. After having a look at the linked doc, I conclude that `getfullargspec` is not actually deprecated (as of Python 3.11). It was deprecated but the decision has been reversed. Still, its use is not recommended and migrating  to signature is the way to go. However, calling it "deprecated" in the comment is misleading, so I kindly ask you to change the comment, e.g. to "Migrate to `signature` as recommended in" (link next line). I am still surprised that it fixes #99 because the linked doc states that `getfullargspec` is already based on `signature` since Python 3.4.
Please feel free to add yourself to the contributors list https://github.com/mdp-toolkit/mdp-toolkit.github.io/blob/master/development.html by making a PR in https://github.com/mdp-toolkit/mdp-toolkit.github.io. (Note that the list is ordered alphabetically by last names)

kotopesutility:
> However, calling it "deprecated" in the comment is misleading, so I kindly ask you to change the comment, e.g. to "Migrate to `signature` as recommended in" (link next line).

Done, I have force-pushed with new comment. Please check it!

PR #103, opened 2023-07-29
FIX: removed dependence from top-level packages on subpackages with tests

kotopesutility:
I think it's better to remove dependencies from top-level packages (mdp, bimdp) on their test subpackages. This will allow you to stop distribute tests with your main package. By the way, it is bad practice to import tests from top-level packages just to run tests.


Stewori:
Thank you again for cleaning this up. Change set looks good and clean, tests are passing I accept it right away.

PR #105, opened 2025-02-17
Port to numpy 2.0

cjwatson:
See https://numpy.org/doc/stable/numpy_2_0_migration_guide.html.


cjwatson:
There are some test failures, but I think they're all from pre-existing problems rather than being caused by this PR.

Stewori:
Thank you, @cjwatson for this PR. We'll have to get the tests green to merge it sanely, but that's a distinct issue. To some extend, it's just about updating the dependency matrix. I'll follow up on this as soon as I can...

